PR #45, opened 2015-04-20
Raising the right error when wrong charcode is used

Krispy2009:
CharCode_error no longer exists and it throws a not defined error when you actually try to raise that exception.
Changing this to CharCodeError seems to fix the problem :)


manpaz:
Thank you Krispy2009!

PR #53, opened 2015-07-04
Extended params for serial printers

ldos:
Updated printer.Serial() constructor and its open() method to extend serial params.
Backwards compatible, no client modifications needed.


manpaz:
Changes proposed makes sense.
Thank you ldos.

Issue #31, opened 2015-03-17
Print image limited to 255px height


GoogleCodeExporter:
```
Hello there,

I am trying to print images that are bigger than 255px in height but it throws 
ImageSizeError(). Is there any reason to limit the image size in hieght ? 

Thanks



```

Original issue reported on code.google.com by `benoit.g...@gmail.com` on 23 May 2014 at 6:06

giorgoc:
Sorry but, since i have the same annoying issue, could you help guide us how to resolve the issue? What limes of code shall i change and where?

Thank you for your precious help!

cod3monk:
How about a simple workaround and split the image into stripes of 255px? I can prepare a patch if you would consider merging it.

PR #59, opened 2015-08-25
Support for images vertically longer than 256 pixels

zouppen:
Fixes issue #31. Tested on Cashino CSN-A3 thermal printer.

Please note that this allows sending huge amounts of data to thermal printer and if you are not using flow control then serial buffers on your printer go nuts and printing gets messy at some point. If you enable CTS/RTS or XON/XOFF then there is no problem at all!

PS. I wrote this patch last February but because you were hosting at Google Code it was not worth the hassle to send it. But since you are on GitHub now and still have that issue, here you are! :-)

### DIFF
--- a/escpos/escpos.py
+++ b/escpos/escpos.py
@@ -41,7 +41,7 @@ class Escpos:
         buffer = ""
        
         self._raw(S_RASTER_N)
-        buffer = "%02X%02X%02X%02X" % (((size[0]/size[1])/8), 0, size[1], 0)
+        buffer = "%02X%02X%02X%02X" % (((size[0]/size[1])/8), 0, size[1]&0xff, size[1]>>8)
         self._raw(buffer.decode('hex'))
         buffer = ""
 
@@ -68,7 +68,7 @@ class Escpos:
 
         if im.size[0] > 512:
             print  ("WARNING: Image is wider than 512 and could be truncated at print time ")
-        if im.size[1] > 255:
+        if im.size[1] > 0xffff:
             raise ImageSizeError()
 
         im_border = self._check_image_size(im.size[0])


matiasw:
Comes with a Works for Me guarantee!

manpaz:
Hey @zouppen ,

Thank you for the contribution, definitive I missed the "x256" multiplier at the height column.
Quick question, to you mind to move change the values to integer instead of HEXs and bitwise operations?
ie:  (size[1]/256) instead size[1]>>8  and 65536 instead 0xffff

I want to give you all the credits for this great fix, and again to thank you for your help.

zouppen:
I can do it of course, but why? Bitwise logic makes sense because ESC/POS data sheet and the underlying serial protocol uses bytes and it is essentially bitwise logic that needs to be done. That's why I see it more self-documenting to reference a run of maximum value in of 16 bit integer as 0xFFFF instead of 65535. And shift of 8 bits (1 byte) is more clear than dividing by magic number. 

Also, representing bitwise logic in decimal numbers is very error prone as you can see when you mixed up 65536 with 0xffff in your message. ;-)

matiasw:
Also, bitshifts are faster than divisions.

matiasw:
Also, bitshifts are faster than divisions.

zouppen:
If you disagree on using bitwise operations, please just rewrite those parts in a subsequent git commit. The most important thing is to get this fixed. :-)

manpaz:
Thank you @zouppen and @matiasw, your contribution is well appreciated.
65535 or "size < 65536" comes from the ESC/POS manual as well as image WIDTHx256 which makes sense why the width should be divided by 256.

Bitshifts are faster than divisions, but have shown no time improvement either on the slowest old Raspberry PI or the faster x86_64 systems.

The idea is to have the code as self-explained as possible, to help the beginners to understand the code, due the lack of comments there, which I feel not proud for not having the time to do it properly :(

In the other hand, thinking it twice, people contributing to the project is usually more knowledgeable, and have no problem to understand the code, so, I'll go the way it is committed, and if and only if necessary I'll made the proper changes ;)

Once again, thank you for your contribution!

taylorstevens:
Boy I picked the right day to download this repo! A few hours after I initially find the issue and it's fixed! Quick question: I'm finding unprinted lines in my images, probably 1px tall, that I assume are caused by the paper feeding too much per line, or in other words the printed areas are not aligning exactly. Is there a way to combat this issue? I'm using an Epson TM-T88V if that helps.

manpaz:
@taylorstevens, all the credits to @zouppen.

About the unprinted lines on your images, do you mind to create a separate ticket please?
Also, I'll appreciate if you can specify the size of the image, and if you have no privacy concerns :)  a download URL could be nice, I'd like to try it on my side.

Thank you for your interest in the project.

matiasw:
It's been opened as Issue #64.

Issue #56, opened 2015-08-05
problem with my printer


dboira:
Hello,
I followed all the steps you describe and I can't write with my printer (FENIX-INVICO).
I get this error message:

pi@raspi2 ~ $ python prn.py
Traceback (most recent call last):
  File "prn.py", line 6, in <module>
    fenix = escpos.Escpos(0x4651,0x0164,0,0x81,0x01)
TypeError: this constructor takes no arguments

This is my prn.py code

#!/usr/bin/env python

from escpos import *

#  """ Seiko Epson Corp. Receipt Printer M129 Definitions (EPSON TM-T88IV) """
fenix = escpos.Escpos(0x4651,0x0164,0,0x81,0x01)
fenix.text("Hello World")
#fenix.image("logo.gif")
#fenix.barcode
#  Epson.barcode('1324354657687','EAN13',64,2,'','')
fenix.cut()

I want to know if it is possible print with this printer.
My sistem is a raspberry pi:
Linux raspi2 3.18.11+ #781 PREEMPT Tue Apr 21 18:02:18 BST 2015 armv6l GNU/Linux

When I put some lsusb command I get a warning message:

pi@raspi2 ~ $ lsusb -vvv -d 4651:0164 | grep iInterface
Couldn't open device, some information will be missing
      iInterface              0

Thanks a lot for your help
Daniel B

manpaz:
HI Daniel,

I'm not sure where you got that example code.
Please, make sure you are following the printer definition described in the wiki:
https://github.com/manpaz/python-escpos/wiki/Usage

For the printer you are using the code to define that printer should be like:
```
fenix = printer.Usb(0x4651,0x0164)
```

Thank you.

dboira:
Hello,
thanks for your response;
previously, I also put that code:

fenix = printer.Usb(0x4651,0x0164)

with the same error.
But today I'll try again.

2015-08-06 0:05 GMT+02:00 manpaz <notifications@github.com>:

> HI Daniel,
>
> I'm not sure where you got that example code.
> Please, make sure you are following the printer definition described in
> the wiki:
> https://github.com/manpaz/python-escpos/wiki/Usage
>
> For the printer you are using the code to define that printer should be
> like:
>
> fenix = printer.Usb(0x4651,0x0164)
>
> Thank you.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/manpaz/python-escpos/issues/56#issuecomment-128163641>
> .
>



-- 
[...]

PR #61, opened 2015-09-04
Prevent crash when using libusb0 printers

HarkonenBade:
libusb0 printers do not have the is_kernel_driver_active method. This patch detects if that method is missing, and if so attempts to detach the kernel driver anyway, ignoring the error if it turns out that the kernel driver is not attached.


HarkonenBade:
Will fix #56

manpaz:
Thank you @HarkonenBade

PR #62, opened 2015-10-09
Updated code for instance creation

diegottl:
This will avoid mislead to a wrong module usage (old usage).


manpaz:
Hi @diegottl ,
Thank you for the contribution, the only thing is that I've planned to remove the usage from the readme as it is outdated.  Anyways, the wiki must supply any usage information.

Once again, thank you.

PR #74, opened 2015-12-26
DOC add very simple (and bad-looking) api-reference as doc-stub

patkan:


patkan:
Merging this in order to check out readthedocs.

PR #87, opened 2016-02-29
Add optional timeout variable for printer network connections

Loafdude:
Small change to add a timeout variable to the socket.


Loafdude:
I set the default timeout to 60 seconds... 
I like mine fairly short but I wasn't sure what to default it to.

patkan:
Thank you!

PR #90, opened 2016-02-29
Add inverted printing option to set method

Loafdude:
Tested working on a TM-T88V.


Loafdude:
Ooops It caught my other commit as well

patkan:
merged, thank you!

PR #93, opened 2016-03-01
Add width/height multiplier support, upsidedown text and text smoothing

Loafdude:
Alright worked a bit on adding some text features.

Firstly upside-down text. Simple enough.
Text smoothing. Effective on text stretched to 4x height or width.

Width/Height.
Escpos has a GS ! command similar to the ESC ! that was currently in use for text size
The GS ! command supports up to 8x scaling of the width and/or height of fonts. 
I do not know what printer support is like for the GS ! vs Esc ! command.
So if the width/height is set to 1 or 2 or combination thereof we use the old method of ESC ! but if height/width > 2 then we use the GS ! method. This maintains full backwards compatibility, sending the same commands down the wire for existing users.

I also fixed a typo in constants.py
There was a comment in the set method that it supported a "B2" text style. 
There is no second bold style I could find supported by the escpos command set.
I removed this line. 

Cheers

### DIFF
--- a/doc/user/methods.rst
+++ b/doc/user/methods.rst
@@ -65,8 +65,8 @@ text("text")
 
 Prints raw text. Raises ``TextError`` exception.
 
-set("align", "font", "type", width, height, invert)
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+set("align", "font", "type", width, height, invert, smooth, flip)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 Set text properties.
 * ``align`` set horizontal position for text, the possible values are:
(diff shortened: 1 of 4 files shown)


patkan:
Looks good to me, thank you!
I checked it with my old T88II and the GS-commands seemed to work. I did not know this.
The only thing that seemed to be ineffective is smoothing, but maybe my test case wasn't good enough.

You are right, I could not find B2 neither, that's a documentation error.

The only thing that could be done is, you could instead of letting pass through the KeyErrors (when someone sets e.g. width=9) throw a self-created error (see exceptions.py and its usage)

Loafdude:
Smoothing is easily visible with width=4 and height=4.
Anything less than that it is barely noticeable. I can confirm it definitely is working on my TM-T88V

I'll look at the error code stuff. 
That's something new to me so I'll see what I can do.

The only other functions I've seen to play with is 90 degree printing and character spacing.
Text functions are fairly complete now

Loafdude:
Alright I think I got it.
Checks for out of range and non-integer fields

Loafdude:
Github got mad at me for the way I was checking values so I changed it
Let me know what you think of the error checking.
I made some assumptions based on the other code.

patkan:
OK, I merge this now, I like it! :+1: Or do you want to directly implement text rotation? Then we could put it with the this stuff in this pull request (simply commiting to you branch will add it to the PR)

The thing with the error codes is just something I took on, when I started working on this project. Maybe we should check someday how other projects do it and what the best practice is (maybe there is a PEP).

By the way: I tried out 4x4 with smoothing. The smoothing my T88II does looks really awkward but it works!

Loafdude:
Yeah the smoothing can look a bit funny

Please merge what is done. I don't need 90 degree printing and am not sure when I will get around to playing with it.

patkan:
OK, thanks again!

PR #94, opened 2016-03-03
Fix problems with python 2 and 3

patkan:
This should fix most python2/3-problems.

Since this change is a bit larger I will keep this PR open a bit before merging.


escpos/printer.py
```diff
@@ -218,11 +226,12 @@ def _raw(self, msg):
...
         :param msg: arbitrary code to be printed
+        :type msg: bytes
         """
-        if type(msg) is str:
-            self.device.write(msg.encode())
-        else:
-            self.device.write(msg)
+        #if type(msg) is str:
+        #    self.device.write(msg.encode())
+        #else:
```
braveheuel: I think it is better to remove the code, not comment it out

patkan: You are totally right.

patkan:
I think this is ready for merging, OK?

Loafdude:
I haven't had any issues running it on python 2.7.x

Issue #85, opened 2016-02-22
python-escpos does not support all barcode types as per esc/pos specification


Loafdude:
Just a heads up some newer printers support more barcode types.
Please see epson's esc/pos reference page here
https://reference.epson-biz.com/modules/ref_escpos/index.php?content_id=128
Sorry I believe it requires registration to access their website.

Specifically 
CODE93
CODE128
GS1-128
GS1 DataBar Omnidirectional
GS1 DataBar Truncated
GS1 DataBar Limited
GS1 DataBar Expanded

patkan:
Yeah you're right. At the moment we are pulling together all the forks out there so there might be large changes in the code which is somewhat unfortunate for Pull Requests, but if you want to you can contribute.
In the long run I would be happy if we could set up something that allows for older printers to emulate those new barcode types by prerendering the barcodes. Maybe we could pass the barcode function simply a flag whether it should send directly to the printer or pre-render.

Issue #58, opened 2015-08-25
No barcode printing with WinBond/Zjiang 5890


matiasw:
Hi,
I'm having trouble with getting barcode printing to work: only the code gets printed, not the bars. My printer is this one: http://www.aliexpress.com/item/Free-shipping-New-mini-58mm-thermal-receipt-printer-ticket-pos-58-thermal-printer-USB-black-white/856027185.html
It is identified as `ATTRS{idVendor}=="0416", ATTRS{idProduct}=="aabb"` and text/image printing work.
Where should I look for the problem? I can see two ways around it: find and fix the issue where barcode is handled, or work around it by rendering the barcode to an image and printing that. I think the first one is simpler, would you agree?

patkan:
Hi @matiasw. Do you know whether the printer speaks "real" esc/pos?
Have you already solved the problem?

bmjbmj:
It took me a while to find the problem. My 5802LD from the same manufacture demands a NUL last in the data field. Python-escpos only sends 'self._raw(code)' NOT self._raw(code+'\0'). A work around is to use a line like: Zjiang.barcode('1324354657687\0','EAN13',64,2,'','') OBS! the '\0'. I'm also confused by "BARCODE_FONT_A  = '\x1d\x66\x00' # Barcode font"  In the Zjiang manual it is described as 'GS f n' and f is not x66. But the Zjiang is a topographical disaster with many errors!

patkan:
You are right. Function type A detects the end of the code with a NUL. Function type B via a supplied length.
So, @DavisGoglin: could you maybe add this to #98 while you're at it? I think we only habe to add a `\x00` in A-mode.

patkan:
OK, I suspect this should be fixed in #98. If no one opposes I will close this issue at the latest after merging into master.

bondzo:
I have this printer [http://www.amazon.de/gp/product/B00LUIST7W/ref=ask_ql_qh_dp_hza](url)
which is also Zjiang ZJ-5890T, ID 0416:5011
Everything works except barcode. I tried both master and development branch. The description says that JAN13 (EAN 13) / JAN8 (EAN8) / CODE39 is supported. The programmer manual even says that all usual barcode types and function type A and B are supported.
I tested the demo-code on #98.
This is what it looks like:
![barcode](https://cloud.githubusercontent.com/assets/19343673/15244521/d39666f0-1901-11e6-8985-84935712e142.jpg)
Looks like something is wrong.

[...]

PR #98, opened 2016-03-07
Adding barcode function type B and additional barocdes. 

DavisGoglin:
Added the option to print barcodes using barcode function type B, requested in issue #85.

I tried to match documentation and test as best I could. I don't have a printer that supports type B. 

As a side note, I'm not sure if we are sending a `NUL` at the end of barcode function type A as is specified in the sec linked.


patkan:
Thank your very much. One thing that I saw: I think it has to be `CODABAR` not `CODEBAR` (https://en.wikipedia.org/wiki/Codabar).
Apart from that: I don't think I have such a printer either. Maybe someone has? @Loafdude, do you have a printer that supports Type B?

Loafdude:
I have a TM-T88V.
From it's specs it should support type B but I cannot get your implementation to work.

That said I can only get type A working with a few of the barcodes.
For example code
    e.text("UPC-A\n")
    e.barcode('132435465768','UPC-A',function_type="A")
    e.text("UPC-E\n")
    e.barcode('127890','UPC-E',function_type="A")
    e.text("EAN13\n")
    e.barcode('1324354657687','EAN13',function_type="A")
    e.text("EAN8\n")
    e.barcode('13243546','EAN8',function_type="A")
    e.text("CODE39\n")
    e.barcode('*TEST*','CODE39',function_type="A")
    e.text("ITF\n")
    e.barcode('55867492279103','ITF',function_type="A")
    e.text("NW7\n")
    e.barcode('31117013206375','NW7',function_type="A")
    e.text("CODE93\n")

Working: UPC-A, EAN13, EAN8, CODE39.
Non-Working: UPC-E, ITF, NW7

When I set the type to "B" the working barcodes no longer work.
They print the barcode 'code' variable unformatted just in standard font. 
No attempt at barcode printing is done. No whitespace, nothing. 

Please add a note to the docs that code39 needs * character at beginning and the end of the data for it to print.

Thanks for working on this.

Loafdude:
test

Loafdude:
Type B on the left, Type A on the right.

Note that the printer in eating some of my text() methods. 
There should be a EAN13 above the second barcode. Type UPC-E is eating that data
Type ITF seems to be eating the N from NW7 text. 

Perhaps the printer is still trying to accept data for these formats.
Not sure on type B. I took a look at the reference but don't have time to debug it.

DavisGoglin:
See if the workaround in issue #58 works to fix any of the type A barcodes - adding a `\x00` at the end of the barcode data.

Loafdude:
That fixed them except the NW7.
That might be a problem with my data though.

\0 should not affect the Function B stuff though. Seems it has a length field in the reference.

DavisGoglin:
If I'm reading the spec right, `e._raw('1d6b650a31323334353637383930'.decode('hex'))` should print a UPC-A barcode of 1234567890. 

- `1D` = `GS` Control character
- `6B` = 'k', barcode command
- `65` = Barcode type UPC-A
- `0A` = 11 characters in the barcode data
- `31323334353637383930` = 1234567890, barcode data

What happens when you print the above?

Loafdude:
Nope thats not working.
Same result as on the left in the image

DavisGoglin:
The only other think I can think of is `e._raw('1d6b410a31323334353637383930'.decode('hex'))`, if 65 is the decimal and we should be sending hex, `41`.

Loafdude:
just tried it no go

Loafdude:
e._raw('1D6B410B3030303030303030303030'.decode('hex'))
works

investigating

Loafdude:
Alright. Here is the two working lengths for UPC-A

e._raw('1D6B410C303030303030303030303030'.decode('hex'))
e._raw('1d6b410b3132333435363738393030'.decode('hex'))

DavisGoglin:
I made a stupid mistake above. I translated `0A` to 11, not 10. So the code I was sending wasn't spec because it was too short of a barcode.

Loafdude:
e.text("UPC-A\n")
    e.barcode('13243546576','UPC-A',function_type="A")
    e.text("UPC-E\n")
    e.barcode('132435','UPC-E',function_type="A")
    e.text("EAN13\n")
    e.barcode('1324354657687','EAN13',function_type="A")
    e.text("EAN8\n")
    e.barcode('13243546','EAN8',function_type="A")
    e.text("CODE39\n")
    e.barcode('*TEST*','CODE39',function_type="A")
    e.text("ITF\n")
    e.barcode('55867492279103','ITF',function_type="A")
    e.text("NW7\n")
    e.barcode('A00000000A','NW7',function_type="A")
    e.text("UPC-A\n")
    e.barcode('13243546576','UPC-A',function_type="B")
    e.text("UPC-E\n")
    e.barcode('132435','UPC-E',function_type="B")
    e.text("EAN13\n")
    e.barcode('1324354657687','EAN13',function_type="B")
    e.text("EAN8\n")
    e.barcode('13243546','EAN8',function_type="B")
    e.text("CODE39\n")
    e.barcode('*TEST*','CODE39',function_type="B")
    e.text("ITF\n")
    e.barcode('55867492279103','ITF',function_type="B")
    e.text("NW7\n")
    e.barcode('A00000000A','NW7',function_type="B")
    e.text("CODE93\n")
    e.barcode('1324354657687','CODE93',function_type="B")
    e.text("CODE128 Type A\n")
    e.barcode('{A*TEST*','CODE128',function_type="B")
    e.text("CODE128 Type B\n")
    e.barcode('{B*TEST*','CODE128',function_type="B")
    e.text("CODE128 Type C\n")
    e.barcode('{C*TEST*','CODE128',function_type="B")
    e.text("GS1-128\n")
    e.barcode('00123456780000000001','GS1_128',function_type="B")
    e.text("GS1 DataBar Omnidirectional\n")
    e.barcode('0000000000000','GS1 DataBar Omnidirectional',function_type="B")
    e.text("GS1 DataBar Truncated\n")
    e.barcode('0000000000000','GS1 DataBar Truncated',function_type="B")
    e.text("GS1 DataBar Limited\n")
    e.barcode('0000000000000','GS1 DataBar Limited',function_type="B")
    e.text("GS1 DataBar Expanded\n")
    e.barcode('00AAAAAAA','GS1 DataBar Expanded',function_type="B")
    e.cut()

test

Loafdude:
Great seems to be working.

GS1-128 is in the docs but your code uses GS1_128 in the dictionary key. Needs a typo fix

Code128 requires setup in the string. We should handle it because it is escpos specific
I would add a CODE128_TYPEA, B and C mode to the dictionary.
append to the end of the string {<mode> before the code (ie. {A, {B, {C)

Other barcodes are quirky. my TM-T88V enforces some requirements on them before it will print. Not all requirements are documented though.
GS1-128 has specific length configuration or something despite the escpos docs saying it can be 2-255. I could not get short strings to print.
GS1-Databar Expanded has a mode set with the first two characters (this is part of spec i believe)
GS1 Databar limited must have a 0 or 1 for the first character (this is part of spec i believe)

Good job. Nice to have that working!

patkan:
I just tested the "demo-code" and those that are supported by my printer (T88II) print fine in A and B mode.

Thank you very much to both of you for your effort.
I think we can merge this as soon as the few minor issues @Loafdude pointed out are fixed.

DavisGoglin:
Added fixes. 

The barcode code above should probably be put into a demo/tests somewhere, but I'll wait until something like fix-image-printing gets merged before I try to figure out how to write this version of tests.

patkan:
OK, looks good to me. I will merge this.

Yes, we should add tests, but I am still not that sure which test framework to use (currently nose, maybe switch to nose2) and how to best write tests for escpos-code. (Probably the best way would be to compose the command as a string from the reference and then compare the output with this reference, but I am open for better propositions.)

DavisGoglin:
In my mind the best method for testing is with you mentioned: get a known good hex output, print to a file, and compare them. 

The only other feature I'd like is to find an easy to way to print to a physical printer for tests.

Issue #42, opened 2015-03-17
Image printing not working on Epson TM U220


GoogleCodeExporter:
```
What steps will reproduce the problem?
1. Connect the printer to the USB bus. The printer is an Epson TM U220
2. Initialize the printer via Epson = printer.Usb instruction 
3. Execute a Epson.image instruction on any kind of image file, format or size

What is the expected output? What do you see instead?

The image should be printed in paper. A large series of ASCII chars are printed 
instead.

What version of the product are you using? On what operating system?


I am using it on Ubuntu 12.04. Text printing works just fine. I have tried 
different image formats, sizes, etc and the error is always the same. 

```

Original issue reported on code.google.com by `Reloadgu...@gmail.com` on 19 Feb 2015 at 10:17

boshkash1151:
I had a similar problem. I solved it by printing and empty line before each image.

theprinter = printer.Usb(0x0000,0x0000)
theprinter.text("\n")
theprinter.image("my_image.jpg")
theprinter.text("\n")


Also png with transparent backgrounds caused an issue. 
Credits go to piniyini (https://github.com/manpaz/python-escpos/issues/51)

mikrogroove:
I'm having the same problem, but with an Epson TM-U295. Looking at constants.py I see that bit-image printing mode is selected with `1D 76 30 00`, while the TM-U295 ESC/POS command reference says it should be `1B 2A 00 00 01` - where the third byte is print mode (TM-U295 only supports 0 or 1 here), fifth is width multiplier (x256), and fourth is width remainder (IIUC). From the TM-U295 manual: 

![epson tm-u295 bit-image mode](https://cloud.githubusercontent.com/assets/196348/8935345/28fff5e0-3544-11e5-8ee2-b3f7fd35e8fe.png)

manpaz:
@mikrogroove , 
Thank you for your interest in the project.
Seems like there has been some issues with transparent images when trying to convert them to RGB. I've committed a fix that hope helps to solve this issue to some many having garbage printed when trying to print RGBA images.

Do you mind to checkout the recent code, do your tests again, and check if you still have the same issues printing images please?

Thank you!

mikrogroove:
[...]

PR #107, opened 2016-03-13
Improve cli.py

DavisGoglin:
### Contributor checklist

 * Star TSP100

----------

### Description

Rewrote `cli.py` to be a mapping of the features of escpos.py.

Usage is `python -m escpos.cli destination_type [--args] escpos_command [--args]`. 

A side effect of using `argparse` is that you can't have more than one subparser, e.g. only destination or only escpos_command. I did a workaround which involves having two instances of `argparse`, each with a single subparser. One runs, then the other runs with whatever commands are left over. As a side effect, the `--help` command doesn't show all of the commands, and adding `--help` shows the help for only destination_type and not escpos_command. Having two subparsers is awkward, but it's nice to get so much functionality just by filling out two `argparse` instances. 

Todo:

 - Add more functionality, such as pre built tests
 - Fix argparse help awkwardness by combining both helps, or change architecture


DavisGoglin:
I moved the printer choice to group arguments so I could use one parser. I had to decide which printer was wanted by checking which arguments were sent, but this is more in the way that argparse was designed.

patkan:
Sorry, I haven't had the time to test this, but it looks good :-)

Just one idea: do you now `ConfigParser`? (I think it is `configparser` in Python 3)
The user could select a printer through a config-file that could be in the home directory, the current directory or similar locations.

DavisGoglin:
It's pretty stupid of me to not have done that in the first place. I still have to write documentation.

It also wouldn't hurt in the future to use the config file elsewhere, it's a pretty simple 50 lines with error handling. I wasn't sure which dir to check in, so I went for a shotgun approach.

patkan:
I haven't used this library yet, but maybe this one is interesting for determining the right directory: https://pypi.python.org/pypi/appdirs

The script could look for the file there and if nothing exists there it could create it from a template.

DavisGoglin:
I split the config loading to it's own module. 

I'm not sure why I need the requirements all to be in `doc/requirements.txt`, but the tests fail if they aren't there.

I still have to write function documentation.

DavisGoglin:
And it seems localconfig breaks python3. Whoops.

patkan:
> I'm not sure why I need the requirements all to be in doc/requirements.txt, but the tests fail if they aren't there.
That is still somewhat messed up. Which tests fail exactly?

DavisGoglin:
I'm not sure what was happening. I was getting import errors, but I'm not getting them now.

Anyway, moved away from localconfig since it was python2 only and switched to yaml storage so that all the datatypes didn't come in as a string. I added docs to all of the functions and made a demo command line argument.

DavisGoglin:
Do you know of an issue open for tests failing on image printing in python3? When running `tox -e py35` I get 

    test the image function with 50x50.png (grayscale png) ... ERROR

Relevant lines:

    pbuffer = "%02X%02X%02X%02X" % (((size[0]/size[1])/8), 0, size[1] & 0xff, size[1] >> 8)
    TypeError: %X format: an integer is required, not float

patkan:
@DavisGoglin I had the import errors too. It seems that something in my tox-environment was stale since after deleting `.tox`, everything was fine...

Considering the problem with py35: I don't know. It seems to me that the `%` formatting might be more strict in 35. But unfortunately I don't have a Python 3.5 at hand at the moment. With Python 34 the tests are all successfull.

I think the best thing would be to rewrite the image printing. I think for this we should wait for the port in #42

patkan:
I just had a look at your work. I like the idea :+1:

Just some issues:
* landscape (in the checks-section of this PR) tells me that there are some unused imports, you should probably remove them
* when I try to load the command-line without config-file the ConfigNotFoundError is thrown. However, in the output there seems to be a newline after every letter of the paths that are suggested)
* You can deploy that script somehow as a commandline tool. I don't know at the moment how, but this would be a nice addition for usability.
* [argcomplete](https://argcomplete.readthedocs.org/en/latest/) would further improve the user experience, however, this could also be added later

DavisGoglin:
If we run into this same issue at rewrite: https://wiki.python.org/moin/PortingToPy3k/34to35#The_.25x_string_format

I'll look into something like using something like [this](https://github.com/offbyone/pybuilder) for testing since I already have docker installed. It covers  2.5 through 3.4, and I have 3.5.1 on arch natively. If you want to test python 3.5 there is always [official docker python](https://hub.docker.com/_/python/). Simple steps are [here](https://civisanalytics.com/blog/engineering/2014/08/14/Using-Docker-to-Run-Python/). 

I was wondering where landscape went. I didn't see the expand button. 

Running it from the command line changes the way imports work. The random SO posts I've found on it all just say to use `python -m escpos.cli` instead. I couldn't find a consistent, nonhacky way to run it differently.

patkan:
> I'll look into something like using something like this for testing since I already have docker installed. 

Good idea! This would make testing probably easier.

> I was wondering where landscape went. I didn't see the expand button.

Landscape is kind of slow. It will sometimes take a bit and only those services that responded will be listed in the checks.

> Running it from the command line changes the way imports work. The random SO posts I've found on it all just say to use python -m escpos.cli instead. I couldn't find a consistent, nonhacky way to run it differently. 

OK, I had no idea. At least for now this is more than sufficient. I will keep in the back of my mind, maybe I stumble upon something useful.

patkan:
Is the demo-function working on your system?
I am using a Serial printer in Python 2.7.
The following error occurs:

```
python -m escpos.cli demo --barcodes-a
Serial printer enabled
Traceback (most recent call last):
  File "/usr/lib/python2.7/runpy.py", line 162, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/usr/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/home/patrick/dev/python-escpos/escpos/cli.py", line 347, in <module>
    main()
  File "/home/patrick/dev/python-escpos/escpos/cli.py", line 325, in main
    globals()[target_command](**command_arguments)
  File "/home/patrick/dev/python-escpos/escpos/cli.py", line 344, in demo
    command(**params)
  File "escpos/escpos.py", line 454, in barcode
    self._raw("\x00")
  File "escpos/printer.py", line 149, in _raw
    self.device.write(msg)
  File "/usr/local/lib/python2.7/dist-packages/serial/serialposix.py", line 518, in write
    d = to_bytes(data)
  File "/usr/local/lib/python2.7/dist-packages/serial/serialutil.py", line 58, in to_bytes
    raise TypeError('unicode strings are not supported, please encode to bytes: %r' % (seq,))
TypeError: unicode strings are not supported, please encode to bytes: u'\x00'
```

When I use Python 3.4 the following error occurs:

```
python3 -m escpos.cli demo --barcodes-a
Traceback (most recent call last):
  File "/usr/lib/python3.4/runpy.py", line 170, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.4/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/patrick/dev/python-escpos/escpos/cli.py", line 347, in <module>
    main()
  File "/home/patrick/dev/python-escpos/escpos/cli.py", line 59, in main
    printer = c.printer()
  File "/home/patrick/dev/python-escpos/escpos/config.py", line 76, in printer
    self.load()
  File "/home/patrick/dev/python-escpos/escpos/config.py", line 46, in load
    if isinstance(config_path, file):
NameError: name 'file' is not defined
```

DavisGoglin:
It works for me on 2.7.11 but I'm printing to File at the moment, and to file, then to network. Does the non-demo barcode work? I'd think it's the same thing.

The python3 error is my fault for not checking that. I don't have a good excuse for that. I assumed file handles worked the same way in both.

DavisGoglin:
It looks like I'm causing problems all around. I wrote 

    if function_type.upper() == "A":
        self._raw("\x00")

In ` def barcode(`, whereas it should be something like 

    self._raw(u"\x00")

Should that be a separate pull request, this one, or will you get that one since you can test it?

patkan:
Oh wow, now I see it. So I am to blame too, I should have seen that when merging :-)
I have pushed a hotfix to development (f76db4e0d48854917325f072c7c7724c714daad3)

(The problem is that a byte-string and not an unicode-string should be used there)

With that fix barcode printing is working.

DavisGoglin:
Also the following:

<code>
-            self._raw(chr(len(code)))
+            self._raw(chr(len(code)).encode())
</code>

I wanted to to images next, but I'll probably be more tests.

patkan:
Are you sure with that last one?

Shouldn't it rather be `self._raw(six.int2byte((len(code)))`? (AFAIR you can't use `chr()` because it's behaviour changed between python 2 and 3)

DavisGoglin:
You're probably right. I did a quick glance at constants to see what to put there.

They both run, but I'm still a little fuzzy between the different string/encoding storage options and python2/3.

patkan:
Stop talking so badly about yourself ;-)
I found https://docs.python.org/3/howto/pyporting.html quite helpful. They also have a section on the difference between string/bytes. I am however still somewhat uncertain when it comes to handling encodings, which we will need for the upcoming release to PyPi.

Do you know rebasing? If you want to include the hotfix from development, you could rebase your branch with `git rebase development` (after you used pull on development). After that your history will not match that in github and so you will have to force-push to your branch on github. (https://www.atlassian.com/git/tutorials/rewriting-history/git-rebase)

DavisGoglin:
It looks like I didn't break everything by rewriting history. I've changed my dev environment to python 3.5 and added pylint so that should help some of the mistakes.

clickworkorange:
> It looks like I didn't break everything by rewriting history.

> “History is a set of lies agreed upon.”
― Napoléon Bonaparte

patkan:
Sorry, I can't give a full diagnosis as I currently don't have the time to look deeply into this, but there still seems to be something defective around barcode printing.
For me, `demo --barcodes-a` prints only the first one. I have tried printing an EAN8 manually, but the printer prints only white space (I have used the EAN8-code from the demo, is that one valid?)

DavisGoglin:
I copied them from python-escpos/pull/98 but it looks like the check digit could have been incorrect. I removed it to allow the printer to add it. I'll be able to check the rest of the barcodes later today.

DavisGoglin:
I ran into a problem where it wasn't sending all of the data before exiting while I was using `Network`. I added a flush before close and that fixed it, so I also added it to some of the other protocols preemptively even though I can't test `Serial`. 

I changed `*TEST*` to `TEST` which fixed CODE39. My printer added the asterisks. I applied the same to CODE128 in hopes that it works the same even though I cant use it. Maybe we can get @Loafdude to see if they CODE39 and CODE128 it work for him in that format.

patkan:
@mike42 wrote in #113:

> I've locally had to adjust cli.py to print a newline: epson.text("Scan to recall TODO list\n"), so that the printer's print buffer is empty before image commands arrive. I'm not including this in the pull request to avoid conflicting with #107

patkan:
Just a quick note since I saw that you are working on that:
I added a coverage analysis. (That is the one check that fails) Ideally a PR should have tests with it, but the cli is somewhat difficult to test. If you want to try something, there exists a library for testing scripts: https://scripttest.readthedocs.org

I think it is more relevant that we concentrate on testing the core functions. So I would say it is no problem, if there are no tests for the CLI. (If we really do not want to test the cli at all, we could excplicitely exclude it)

Just put a note here when you are done and I will merge this into development.

DavisGoglin:
I've got a quick couple things to finish up them I'm happy with it. Once I finish wrestling with vim and it's regex.

DavisGoglin:
- I think 8101e1ec9f helps maintainability, but I'm open to input. 
- I added a couple tests just to start a structure for it, but you're right in saying they'll provide more value testing the library directly. 
- I removed the executable flag from cli.py since it can't be directly executed at the moment, and it was causing code coverage to skip it. Not that it has great coverage anyway... :)

Unless there is anything else, I'm finished with this for now.

test/test_cli.py
```diff
@@ -0,0 +1,102 @@
+"""Test for the CLI
+
+"""
+
```
patkan: You should probably include

```python
from __future__ import absolute_import
from __future__ import division
from __future__ import print_function
from __future__ import unicode_literals
```
here. Just to be sure that we do not introduce any problematic behaviours without realising it.

test/test_cli.py
```diff
@@ -0,0 +1,102 @@
...
+    devfile: {testfile}
+'''.format(
+    testfile=DEVFILE,
+)
+
+class TestCLI:
+    """ Contains setups, teardowns, and tests for CLI
+    """
+
+    @classmethod
```
patkan: QuantifiedCode found this line. Is there any cause that you do not use a staticmethod?
Could you please change that to `@staticmethod` (and remove the `cls` in `(cls)`)?

test/test_cli.py
```diff
@@ -0,0 +1,102 @@
...
+    def setup_class(cls):
+        """ Create a config file to read from """
+        with open(CONFIGFILE, 'w') as config:
+            config.write(CONFIG_YAML)
+
+    @classmethod
```
patkan: Same here.

patkan:
8101e1ec9f9a2bfa7f930b2f4124e4859a4c48ae looks like a good idea to me.

I have commented on some things but apart from those small issues I would merge it like this. Could you please look into them and then I will merge at the weekend so that there is still time if someone else wants to have a look at it.

escpos/cli.py
```diff
@@ -1,213 +1,515 @@
...
-    """Wrapped in a function for import and entry point compatibility"""
-    # pylint: disable=bad-continuation
-
-    import argparse
+"""
 
-    parser = argparse.ArgumentParser(
-        description="Command-line interface to python-escpos")
-    subparsers = parser.add_subparsers(title='subcommands')
+from __future__ import absolute_import
```
patkan: It probably does no harm to include all 4 `__future__`-imports.

escpos/config.py
```diff
@@ -0,0 +1,84 @@
+""" ESC/POS configuration manager.
+
+This module contains the implentations of abstract base class :py:class:`Config`.
+
+"""
+
+from __future__ import absolute_import
```
patkan: Same here.

DavisGoglin:
I could have sworn I tried to change those `@classmethod`s before and it didn't work. It's working now.

I've made the above suggested changes.

escpos/config.py
```diff
@@ -0,0 +1,89 @@
...
+
+        """
+        if not config_path:
+            config_path = os.path.join(
+                appdirs.user_config_dir(self._app_name),
+                self._config_file
+            )
+
+        try:
+            if isinstance(config_path, str):
```
patkan: Sorry, I have not seen that one before, but by importing unicode_literals I have found a problem:
In Python 2 the type of `config_path` is (then) unicode. That's why the config is not properly loaded and therefore the else-clause is taken. (I don't quite understand that one; why are you calling yaml.safe_load with config_path?)
The easiest fix to this type-check is to write `isinstance(config_path, (str, unicode))` instead of `isinstance(config_path, str)`. 

One further idea: Maybe you should throw an exception if the `printer`-section hasn't been found in the config. Otherwise the resulting error (1) will be rather unintuitive. This can however also be fixed later.

(1)
```
self._printer = getattr(printer, self._printer_name)(**self._printer_config)
TypeError: getattr(): attribute name must be string
```

DavisGoglin:
I'm confused as to how it passed the tests? I removed my local configuration so if wasn't getting an alterative config path, it should have failed?

I'm checking if it was a string so it could handle loading a passed file path, or a passed file handle. If pyyaml gets a string it assumes it's yaml so I need to read the file first. That part obviously requires some comments.

I wasn't aware that `is_instance` worked like that. I'll be able to make that part much clearer by checking for a file handle in both python 2 and 3. 

I even have a comment in there about throwing an error but I decided not to for some reason. It probably makes more sense to do that than `return None`, right?

DavisGoglin:
I accidentally clumped together commits, but b9c9189 hs the py2 unicode fix. I check for a file type object first to see if I can give it to pyyaml, otherwise try to open it first. 

Eventually I need to write tests for `escpos.config`

patkan:
Thank you for the quick fixes!
I have found a bug that is hard to spot: boolean options do not seem to work in argparse. That's because whatever you pass will always evaluate to True by the way casting works in Python.
This seems to affect all boolean options in the cli (By the way, I really like the structure you used for the cli-definition. Makes it much more maintainable.)

Here is a Stackoverflow describing another method: http://stackoverflow.com/a/31347222/4244236
If you do not have the time to fix it now, I would put up an Issue so we can fix it later. (The CLI is atm nevertheless usable, even without this fix.)

DavisGoglin:
Actually, disregard the above commit.

DavisGoglin:
I wanted to go with your suggested method, but after I tried implementing it it has a side effect. If a bool value isn't passed it assumes it's set to false. This could cause problems when a method has an argument that defaults to true. 

    python -m escpos.cli set --no-flip
    {'invert': False, 'smooth': False, 'flip': False}

    python -m escpos.cli set --flip
    {'invert': False, 'smooth': False, 'flip': True}


If it's called from python, it would default to true, but from the CLI it would default to false. I think fca3631 is the better way to go since it only send true or false if the user explicitly passes the argument. This preserves the method's default values.

patkan:
Your solution looks good and works for me.
I think we can merge that now, ok?

DavisGoglin:
Poor code coverage, but yeah, I'm done here.

patkan:

PR #111, opened 2016-03-19
Clean up non-float value, add v3.5 to tox test

mike42:
One of the divisions should be returning an integer rather than a float. This change adjusts a division operator and adds python 3.5 to the tox.ini file.

Note that the discussion on #107 refers to the same problem, and that test failures in #110 log this line as the culprit, as the `.format()` syntax it uses picks up on the float value.

    pbuffer = "{0:02X}{1:02X}{2:02X}{3:02X}".format(((size[0]/size[1])/8), 0, size[1] & 0xff, size[1] >> 8)
    nose.proxy.ValueError: Unknown format code 'X' for object of type 'float'

Tested by printing the test images after the change, to an Epson TM-T20II.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,14 @@
 language: python
 sudo: false
 cache: pip
+matrix:
+  include:
+  - python: 2.7
+    env: TOXENV=py27
+  - python: 3.4
+    env: TOXENV=py34
+  - python: 3.5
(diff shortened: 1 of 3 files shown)


patkan:
Thank you!
Looks good to me and works for me too, I'll merge it.

Issue #22, opened 2015-03-17
QR slow


GoogleCodeExporter:
```
Hi, i'm using this library for raspberry and everything works fine.

I've got a "problem" when trying to print QR codes with an url content.
The problem is that is too slow, like 10 or 20 seconds to process it, and all 
the system is waiting for finish it. I know that i can use background command 
but i need to wait until it finish.

Is this normal? Also is too slow printing images.
I've used all the library packages on the install guide.

Thanks!

PD: Sorry for my English!


```

Original issue reported on code.google.com by `j.doming...@kodeline.com` on 2 Mar 2014 at 12:07

PR #113, opened 2016-03-20
Implement ESC/POS QR code rendering

mike42:
This pull request implements the ESC/POS command for QR code rendering: `GS ( k`. The added code is a direct port of the same feature in in [mike42/escpos-php](https://github.com/mike42/escpos-php) (see [Printer.php](https://github.com/mike42/escpos-php/blob/550d7bc5b5c6c81cea42099bc51a0fbf7c797b87/src/Mike42/Escpos/Printer.php), [EscposTest.php](https://github.com/mike42/escpos-php/blob/550d7bc5b5c6c81cea42099bc51a0fbf7c797b87/test/unit/EscposTest.php), mike42/escpos-php#19). As in the original code, the error correction level, code model and block size are now user-selectable, and test cases have also been ported. This was tested on an Epson TM-T20II.

On printers that don't support the command, existing code wont break, but some new (optional) parameters are now added-

```python
epson.qr("http://www.example.com/", size=10)
```

```
$ time python -m escpos.cli test

real	0m1.852s
user	0m1.056s
sys	0m0.700s
```

But on printers with `GS ( k`, the QR code can be rendered on the printer instead. The `escpos.qr` method simply needs `native=True` to render the same output on the printer:

```python
epson.qr("http://www.example.com/", native=True, size=10)
```

```
$ time python -m escpos.cli test

real	0m0.073s
user	0m0.052s
sys	0m0.008s
```

Because image processing is not used, the second example executed some 25x faster, probably fixing #22.

--

Notes-
- I've locally had to adjust `cli.py` to print a newline: `epson.text("Scan to recall TODO list\n")`, so that the printer's print buffer is empty before image commands arrive. I'm not including this in the pull request to avoid conflicting with #107
- Feedback on code style/quality and/or adherence to project conventions would be appreciated, as I'll be porting some more substantial code in the next few weeks.

### DIFF
--- a/escpos/constants.py
+++ b/escpos/constants.py
@@ -183,6 +183,16 @@ BARCODE_TYPES = {
     'B': BARCODE_TYPE_B,
 }
 
+## QRCode error correction levels
+QR_ECLEVEL_L = 0;
+QR_ECLEVEL_M = 1;
+QR_ECLEVEL_Q = 2;
+QR_ECLEVEL_H = 3;
+    
+## QRcode models
+QR_MODEL_1 = 1;
+QR_MODEL_2 = 2;
+QR_MICRO = 3;
 
 # Image format
 # NOTE: _PRINT_RASTER_IMG is the obsolete ESC/POS "print raster bit image"
--- a/escpos/escpos.py
+++ b/escpos/escpos.py
@@ -253,22 +253,85 @@ class Escpos(object):
         self._raw(binascii.unhexlify(bytes(buf, "ascii")))
         self._raw(b'\n')
 
-    def qr(self, text):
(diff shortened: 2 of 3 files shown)


patkan:
Thank you! Unfortunately I can't test it on my TM-T88II, but the code looks good.
I should probably put down some code-style-guidelines. (#114)
However, your style looks good to me and is easy to read. I very much like that you provide comments. :-)
I would say go on like this.

Just one thing: you should adapt the comment-section in the top of the tests file.

mike42:
Cheers, I'll close this and open a new pull request with the changes, since there's a rebase in there.

PR #116, opened 2016-03-26
improve travis configuration with more versions

patkan:
Testing compatibility: add python 3.3, pypy, pypy3, 3.5-dev, 3.6-nightly to .travis.yml
allow failure on pypy3, 3.5-dev and nightly

@mike42: I just found the rest off your branch and liked it. I have rebased it to current development HEAD. I hope that's OK for you.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,12 +3,26 @@ sudo: false
 cache: pip
 matrix:
   include:
-  - python: 2.7
-    env: TOXENV=py27
-  - python: 3.4
-    env: TOXENV=py34
-  - python: 3.5
-    env: TOXENV=py35
+    - python: 2.7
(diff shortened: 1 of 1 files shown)


mike42:
Glad you find it useful. This configuration will get you red flag in Travis CI if you introduce compatibility issues on a version that used to work.

I'll note here that the tests don't pass on pypy3 and python 3.6-dev. This has to do with some old-style string formatting and a `nose` bug respectively.

image

--

patkan:
> I'll note here that the tests don't pass on pypy3 and python 3.6-dev. This has to do with some old-style string formatting and a nose bug respectively.

Thank you, I noticed that too and I think this is a good addition. This way we can see that there are issues that have to be fixed in future versions.

PR #125, opened 2016-04-03
Implement ESC/POS QR code rendering

mike42:
This is a new pull request based on feedback at #113

> This pull request implements the ESC/POS command for QR code rendering: GS ( k. The added code is a direct port of the same feature in in mike42/escpos-php (see Printer.php, EscposTest.php, mike42/escpos-php#19). As in the original code, the error correction level, code model and block size are now user-selectable, and test cases have also been ported. This was tested on an Epson TM-T20II.

The changes are-

- Rebased to head of `development`
- File header in `test_function_qr_native.py` has been updated.

### DIFF
--- a/escpos/constants.py
+++ b/escpos/constants.py
@@ -183,6 +183,16 @@ BARCODE_TYPES = {
     'B': BARCODE_TYPE_B,
 }
 
+## QRCode error correction levels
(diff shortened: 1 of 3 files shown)


patkan:
Thank you!

Issue #132, opened 2016-04-28
change all references to readthedocs


patkan:
readthedocs has moved to readthedocs.io with the hosted content. They are forwarding, but we should change this in the near future.

PR #133, opened 2016-04-30
Converted references to readthedocs.org -> readthedocs.io

mike42:
Used a find and replace to update the hosted content URLs that are currently redirecting.

    grep -irl --exclude-dir='.*' "readthedocs\." | xargs sed -i 's/python-escpos.readthedocs.org/python-escpos.readthedocs.io/g'

Unless I've missed something obvious, this should fix python-escpos/python-escpos#132.


patkan:
Thank you, that seems to be all.

PR #135, opened 2016-05-01
DOC add example on printing commands from binary files

patkan:
I have added a section to the documentation about printing directly from binary.

@mike42 is it ok to include your barcode.bin as an example?


patkan:
Thank you!

PR #136, opened 2016-06-17
Feature/mosquito compatibility

patkan:
This should add the with-statement and very basic codepage-encoding.


mosquito:
Excellent

PR #149, opened 2016-07-24
Improve/text encoding

patkan:
### Contributor checklist

 * e.g. Epson TM-T88II

----------

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,9 +6,12 @@ Changelog
 
 changes
 ^^^^^^^
+- feature: the driver tries now to guess the appropriate codepage and sets it automatically
+- as an alternative you can force the codepage with the old API
 
 contributors
 ^^^^^^^^^^^^
+- Patrick Kanzler (with code by Frédéric Van der Essen)
 
 
 2016-08-02 - Version 2.1.1 - "Contents May Differ"
--- a/doc/user/usage.rst
+++ b/doc/user/usage.rst
@@ -177,6 +177,20 @@ And for a network printer::
                 host: 127.0.0.1
                 port: 9000
 
+Printing text right
+-------------------
+Python-escpos is designed to accept unicode. So make sure that you use ``u'strings'`` or import ``unicode_literals``
+from ``__future__`` if you are on Python2. On Version 3 you should be fine.
+
+For normal usage you can simply pass your text to the printers ``text()``-function. It will automatically guess
+the right codepage and then send the encoded data to the printer. If this feature should not work, please try to
+isolate the error and then create an issue.
+
+I you want or need to you can manually set the codepage. For this please use the ``charcode()``-function. You can set
+any key-value that is in ``CHARCODE``. If something is wrong, an ``CharCodeError`` will be raised.
+After you have set the codepage manually the printer won't change it anymore. You can get back to normal behaviour
+by setting charcode to ``AUTO``.
+
 Advanced Usage: Print from binary blob
 --------------------------------------
 
@@ -204,19 +218,3 @@ Here you can download an example, that will print a set of common barcodes:
 
     * :download:`barcode.bin </download/barcode.bin>` by `@mike42 <https://github.com/mike42>`_
 
-How to update your code for USB printers
-----------------------------------------
-
-Old code
-
-::
-
-    Epson = escpos.Escpos(0x04b8,0x0202,0)
-
-New code
-
-::
-
-    Epson = printer.Usb(0x04b8,0x0202)
-
-Nothe that "0" which is the interface number is no longer needed.
--- a/src/escpos/constants.py
+++ b/src/escpos/constants.py
@@ -106,7 +106,7 @@ CHARCODE = {
     'PC437':
         [ESC + b'\x74\x00', 'cp437'],  # PC437 USA
     'KATAKANA':
-        [ESC + b'\x74\x01', 'katakana'],  # KATAKANA (JAPAN)
+        [ESC + b'\x74\x01', ''],  # KATAKANA (JAPAN)
     'PC850':
         [ESC + b'\x74\x02', 'cp850'],  # PC850 Multilingual
     'PC860':
--- a/src/escpos/constants.py
+++ b/src/escpos/constants.py
@@ -102,26 +102,98 @@ TXT_INVERT_ON  = GS  + b'\x42\x01'  # Inverse Printing ON
 TXT_INVERT_OFF = GS  + b'\x42\x00'  # Inverse Printing OFF
 
 # Char code table
-CHARCODE_PC437  = ESC + b'\x74\x00'  # USA: Standard Europe
-CHARCODE_JIS    = ESC + b'\x74\x01'  # Japanese Katakana
-CHARCODE_PC850  = ESC + b'\x74\x02'  # Multilingual
-CHARCODE_PC860  = ESC + b'\x74\x03'  # Portuguese
-CHARCODE_PC863  = ESC + b'\x74\x04'  # Canadian-French
-CHARCODE_PC865  = ESC + b'\x74\x05'  # Nordic
-CHARCODE_WEU    = ESC + b'\x74\x06'  # Simplified Kanji, Hirakana
-CHARCODE_GREEK  = ESC + b'\x74\x07'  # Simplified Kanji
-CHARCODE_HEBREW = ESC + b'\x74\x08'  # Simplified Kanji
-CHARCODE_PC1252 = ESC + b'\x74\x11'  # Western European Windows Code Set
-CHARCODE_PC866  = ESC + b'\x74\x12'  # Cirillic #2
-CHARCODE_PC852  = ESC + b'\x74\x13'  # Latin 2
-CHARCODE_PC858  = ESC + b'\x74\x14'  # Euro
-CHARCODE_THAI42 = ESC + b'\x74\x15'  # Thai character code 42
-CHARCODE_THAI11 = ESC + b'\x74\x16'  # Thai character code 11
-CHARCODE_THAI13 = ESC + b'\x74\x17'  # Thai character code 13
-CHARCODE_THAI14 = ESC + b'\x74\x18'  # Thai character code 14
-CHARCODE_THAI16 = ESC + b'\x74\x19'  # Thai character code 16
-CHARCODE_THAI17 = ESC + b'\x74\x1a'  # Thai character code 17
-CHARCODE_THAI18 = ESC + b'\x74\x1b'  # Thai character code 18
+CHARCODE = {
+    'PC437':
+        [ESC + b'\x74\x00', 'cp437'],  # PC437 USA
+    'KATAKANA':
+        [ESC + b'\x74\x01', 'katakana'],  # KATAKANA (JAPAN)
+    'PC850':
+        [ESC + b'\x74\x02', 'cp850'],  # PC850 Multilingual
+    'PC860':
+        [ESC + b'\x74\x03', 'cp860'],  # PC860 Portuguese
+    'PC863':
+        [ESC + b'\x74\x04', 'cp863'],  # PC863 Canadian-French
+    'PC865':
+        [ESC + b'\x74\x05', 'cp865'],  # PC865 Nordic
+    'KANJI6':
+        [ESC + b'\x74\x06', ''],  # One-pass Kanji, Hiragana
+    'KANJI7':
+        [ESC + b'\x74\x07', ''],  # One-pass Kanji
+    'KANJI8':
+        [ESC + b'\x74\x08', ''],  # One-pass Kanji
+    'PC851':
+        [ESC + b'\x74\x0b', 'cp851'],  # PC851 Greek
+    'PC853':
+        [ESC + b'\x74\x0c', 'cp853'],  # PC853 Turkish
+    'PC857':
+        [ESC + b'\x74\x0d', 'cp857'],  # PC857 Turkish
+    'PC737':
+        [ESC + b'\x74\x0e', 'cp737'],  # PC737 Greek
+    '8859_7':
+        [ESC + b'\x74\x0f', 'iso8859_7'],  # ISO8859-7 Greek
+    'WPC1252':
+        [ESC + b'\x74\x10', 'cp1252'],  # WPC1252
+    'PC866':
+        [ESC + b'\x74\x11', 'cp866'],  # PC866 Cyrillic #2
+    'PC852':
+        [ESC + b'\x74\x12', 'cp852'],  # PC852 Latin2
+    'PC858':
+        [ESC + b'\x74\x13', 'cp858'],  # PC858 Euro
+    'KU42':
+        [ESC + b'\x74\x14', ''],  # KU42 Thai
+    'TIS11':
+        [ESC + b'\x74\x15', ''],  # TIS11 Thai
+    'TIS18':
+        [ESC + b'\x74\x1a', ''],  # TIS18 Thai
+    'TCVN3':
+        [ESC + b'\x74\x1e', ''],  # TCVN3 Vietnamese
+    'TCVN3B':
+        [ESC + b'\x74\x1f', ''],  # TCVN3 Vietnamese
+    'PC720':
+        [ESC + b'\x74\x20', 'cp720'],  # PC720 Arabic
+    'WPC775':
+        [ESC + b'\x74\x21', ''],  # WPC775 Baltic Rim
+    'PC855':
+        [ESC + b'\x74\x22', 'cp855'],  # PC855 Cyrillic
+    'PC861':
+        [ESC + b'\x74\x23', 'cp861'],  # PC861 Icelandic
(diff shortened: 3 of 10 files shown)


mike42:
Hey,

Chipping in with how I've implemented code page handling in this project's PHP sibling.

I've got these languages working: Danish, German, Greek, English, Spanish, French, Irish Gaelic, Hungarian, Icelandic, Japanese (Katakana only at this stage), Latvian, Polish, Russian, Turkish and Vietnamese. Feel free to adapt my unit tests if this approach sounds useful to you.

This approach is not sufficient for scripts that don't fit in half a code page. Users have had mixed results with Thai, and Chinese support is via a printer-specific hack, and I was unable to get the contextual forms of Arabic letters to display. In addition, Hebrew and Arabic also print backwards, Japanese support is incomplete (printer issues), and Indian scripts have not been attempted at all.

### Data representation

When printing, the driver needs to know:

- Which code pages I can find characters in. This is stored as a map of UTF-8 characters to lists of code page numbers (called `$available`)
- How to convert the characters to the target code page. For each code page, this is stored as a map of UTF-8 characters to the characters in the target code page (called `$encode`)

The data sources here are:

- `iconv` output, allowing checks for what's in each code page
- The printer model properties, containing lists of code pages ([example](https://github.com/mike42/escpos-php/blob/f6bb510e6f3668ae278d7aec3afb029104e1459f/src/Mike42/Escpos/CapabilityProfiles/DefaultCapabilityProfile.php), [other example](https://github.com/mike42/escpos-php/blob/cfea4c4fc95516ac953e1e5b623f854632afa2ee/src/Mike42/Escpos/CapabilityProfiles/StarCapabilityProfile.php))

### Algorithm

- The driver reads as long as it can find that the characters exist in the current code page (checking that the current code page is on the `$available` list for the character).
- When it finds something it can't encode, it adds a "?".
- It puts the text through a conversion from UTF-8 to the current code page (using the `$encode` table for the code page), and sends it off to the printer.
- When it finds something in another code page, it changes code page and repeats.

### Implementation

The implementation of building this data structure and working with it is:

- [EscposPrintBuffer.php](https://github.com/mike42/escpos-php/blob/cfea4c4fc95516ac953e1e5b623f854632afa2ee/src/Mike42/Escpos/PrintBuffers/EscposPrintBuffer.php)

The unit tests are more interesting though:

- [EscposPrintBufferTest.php](https://github.com/mike42/escpos-php/blob/f6bb510e6f3668ae278d7aec3afb029104e1459f/test/unit/EscposPrintBufferTest.php)

patkan:
OK, thank you! Maybe I can salavage something from your code since the basic strategy seems to be very similar.
Just one point: can I really be sure which escape sequences will be in the output? I mean, the algorithm could find another codepage before the one that we assumed in the unittest. Maybe the better variant would be to decode the output into unicode again (and stripping all control sequences). Then we can compare two unicode strings. The only thing that then still has to be handled are unprintable symbols.

miracle2k:
I need this too and would be happy to help getting it merged.

patkan:
Cool, what do you think of the basic approach? (automatically choosing codepages)
I thought of maybe implementing something like a Dijkstra in order to search the longest usable encoding. On the other hand this is probably complete overkill. Changing codepages costs near to nothing.

miracle2k:
I'm primarily interested in seeing a working version being in the development branch. Later, the search algorithm can be changed if sensible.

I see that @mike42 PHP version is pregenerating and caching the maps in two structures like this:

```
available = {
  'unicodeCodePoint' = {
    'codepageNumberInESC': byte
  }
}

encoded = {
  'codepageNumberInESC': {
    'unicodeCodePoint': byte
  }
}
```

It's a good idea. Not searching all encodings for every character as we do now would definitely improve speed. A compromise might be to generate the cache on the fly.

I'm pretty new to ESC/POS printing and I'm still figuring stuff out. What I noticed, attempting to print this damn Euro sign, is that this algorithm is not enough. For a Euro sign it selects `WPC1251`, which my TM88IIIP doesn't support. The euro sign there is in 858.

So I think we can't avoid having some kind of capabilities system, where we can say "this code page is not available".

patkan:
Sounds reasonable. For the stuff that I implemented I planned to add a list of "allowed" codepages. This way at first the user and later the capabilities-system could block certain codepages.

When we create caches the most elegant way would be probably by using and lrucache. There is a python-lib that allows adding a lrucache-decorator to a function. Whenever you call a function with the same arguments the cache is automatically called. (You thus don't need to interact with the cache)

miracle2k:
My suggestion for how to define the codepages via capability profiles is this:

https://github.com/miracle2k/python-escpos/blob/3fd1a3de5d2824d8c60544ce4b967c5d8391179b/src/escpos/capabilities.yml

The data is mostly from the php project, and the testing/docs from my own printer.

mike42:
I've got some time to prototype changing the PHP driver's encoding/compatibility features to use static data files in the next few days. I'll be working in [mike42/escpos-snippets](https://github.com/mike42/escpos-snippets). If I can drop the PHP code, I'd be glad to start storing printer compatibility info from a repo shared by open source receipt printing projects, to minimize duplication of effort.

I've had a bit of a think about the data layout, and what I've come up with looks slightly different to this pull request, but with the same features. I think that if we store the data like this, it gives us more usage scenarios. Eg, we could use the data for escpos-php, python-escpos, a parser, render to a web-page, or whatever else.

Can you let me know if you see any issues if I generated YAML files in the format below as a part of this?

The three YAML files model:

- Profiles
- CodePages
- Commands

## Profile

Needs to capture all of the data that you get from a test print on a given printer. Included some free-text fields so that we don't need to comment the file, and a printer list so that [printer lists](https://github.com/mike42/escpos-php#printers) can include the best known profile to use.

```yaml
---
default:
  name: Default profile
  description: Default ESC/POS profile for Epson-branded printers
  printers:
  - FooPrint 90000
  - SuperPrint 80
  command:
    LineFeedCmd: false
    FeedAndCutCmd: true
    SetAbsolutePrintPosCmd: true
    GraphicsDataCmd: true
  media:
    width:
      pixels: 768
      mm: 80
  font:
    0:
      columns: 40
    1:
      columns: 50
  codepage:
    0: CP437
    7: TCVN-3-1
  colors: 2
  features:
    starCommands: false

simple:
  name: Simple profile
  inherit: default
  description: Simple profile for printers which do not support more advanced features
  command:
    GraphicsDataCmd: false
...
```

## CodePage

The upper half of the code page (chars 128-255). Encodings are sourced from `iconv`, `custom`, or `none` (stuff we know exists but we can't encode).

I'm suggesting that we pull custom code pages out of the profiles and put them here, as it's really just another data source.

My ideal `codepage.yml` would look like:
```yaml
---
codepage:
  CP437:
    name: OEM font
    provider: iconv
  CP1098:
    name: CP1098 Farsi
    provider: none
  TCVN-3-1:
    name: TCVN-3 Vietnamese 1
    provider: custom
    data:
      - "                "
      - "                "
      - "        ăâêôơưđ "
      - "     àảãáạ ằẳẵắ "
      - "      ặầẩẫấậè ẻẽ"
      - "éẹềểễếệìỉ   ĩíịò"
      - " ỏõóọồổỗốộờởỡớợù"
      - " ủũúụừửữứựỳỷỹýỵ "
...
```

## Command

I've got a parser in the works which has to identify ESC/POS commands, so I have given them all unique names.

Including the command-list will let us query support for a given command, to perform substitutions or disable functionality based on the printer that is connected.

```yaml
---
command:
  LineFeedCmd:
    name: Line feed
  FeedAndCutCmd:
    name: Feed and cut
  SetAbsolutePrintPosCmd:
    name: Set absolute print position
  GraphicsDataCmd:
    name: Graphics data
...
```

miracle2k:
Some good ideas, here are my thoughts:

- colors should be listed, I think, as actual colors (black, red, or something else). There isn't really any cost to collecting this information.

- I don't know about the `provider` key of codepages. `iconv` is platform specific. In Python, we'd just use their builtin encoding engine (which may are may not use iconv, I don't know). The point being, these encodings are standardized, and implementors are free to convert them however they like. Non-standard encodings we can provide with a `data` key. If there is no `data` and the implementor doesn't have any other source, he can't encode. So no need for `none` either. I would just allow an optional `data` key.

- In designing my attempt at a YAML format, I noticed that some printer have different column configurations. TM-P80 has a "42 column emulation mode", and TM-P60II 2's columns depend on the paper size. My approach was to define the different "column configs" and then basically in code letting the user pass some arguments to the profile, i.e. TM80Profile(emulation_mode=True). In any case, there should be some way we can store this information.

  I think an idea I like much more now is to simply define - for those cases - two profiles. "TM-P80 (42-column emulation)" could be that.

That brings me to the last point. I'm not sure about the `printers` list. I gather you want to define not a profile for each printer, like I did, but have base profiles that link to the printers.

Now you know the printer ecosystem better than me, but my impression at least is that there are bound to be a bunch of models out there that have the same command set, but may differ in columns or media size. So I think we might end up with something like this:

     TM80Profile:
         printers:
             - TM80

On a deeper level, I am worried that this format encourages us to attempt to merge duplicate information in shared profiles that really should be duplicated and stored with the printer model. In other words, if there are three printers with the same `media/width` configuration, I would argue that this should be written out three separate times. There should not be a `Foo` profile that tries to accommodate a set of printer models that are not really related, except that their meda size happens to be the same. 

I'm not sure if I'm explaining this right. I'm afraid the profiles and printers get muddled up.

For example, I don't think the `default` profile should contain any information about columns or media size. As an implementor,  should be able to tell, from the data, whether the column or width information I have is just a fallback default, or actual information.

I imagine that printers could be a first class data structure; they do "inherit" from profiles in the sense that both can contain the same keys. Maybe the printer would allow some extra information like the manufacturer to be stored.

     printers:
        TM80P:
             inherits: default
             media:
                width:
                    mm: 72

miracle2k:
One more thing I'd like to see is aliases for codepages. CP437 for example should be addressable as "PC437" or "USA" too.

miracle2k:
Using @mike42's suggestions this I changed the YAML to this:

https://github.com/miracle2k/python-escpos/blob/216184f43f2080520df7c6ab541f80d382471842/src/escpos/capabilities.yml

Notes:

1. I'm still struggling with the code pages. Why do we need to list them at all?

For example, for this Python implementation, during testing the encoder tried to use "RK1048" , which Python couldn't encode. Turns out they added that in Python 3.5. For PHP, I might compile my iconv to include some of the encodings that are currently disabled in the PHP driver. 

So again, the encodings are standardized, and what every implementation can support is up to them, and might change depending on the platform.

What good is it to list the code pages at all? Relying on the platform's encoding names works for 90% of the use cases. Except for one thing:

If my printer profile lists "I support cp437", but my user is used to providing the code page using just "437", because the platform's codecs system supports that alias, how can we find out that this encoding is supported by the chosen printer profile - since the strings don't match?

Should we, or shouldn't we, build our own "database" of encoding names and their aliases? If we don't, it's up to each implementation to either:

a) require the user to specify precisely the encoding strings that the profile uses.
b) use some other data source for encoding alias names.

NB: This only is relevant for the case that the user wants to specify an encoding manually. As long as our implementations use a "auto encode" system, the only thing that is important is that the codec system each implementation uses can resolve the "common encoding" name that the profile specifies. My guess is that we can assume that they all can deal with a name like "cp437".


2. I went on and on above about not putting an "approximate media info" or "approximate column count" inside the basic profiles, but I did here anyway. The default profile contains two fonts and column counts.

- It means that an implementor can use/expose this profiles "as-is", in the same way they would use the printer profile, and can rely on having column information, rather than having to provide their own defaults.

- By separating the base profiles and the printer definitions into two separate dicts, it's still possible to tell which information is certain knowledge (if I have a printer), and which is a guess (when I have a profile).


I'm curious what both @patkan  and @mike42 think. I'm looking forward to seeing this into the library.

mike42:
I will try to address this properly in code, but two specific points-

> So again, the encodings are standardized, and what every implementation can support is up to them, and might change depending on the platform.

mike42/escpos-php#62 is a counter-example for the expectation that code pages are consistently implemented. In any case, legacy code pages are very niche, so I think we will need to be prepared to bridge the gap between what the computer speaks, and what the printer understands.

I used to see a lot of bug report traffic from Windows users who were struggling to encode. An encoding cache is now bundled with the escpos-php library, so `iconv` is no longer a run-time dependency. I think this driver could eventually introduce something similar if it becomes a problem.

> What good is it to list the code pages at all? Relying on the platform's encoding names works for 90% of the use cases

There are different responsibilities between the profiles (which map numbers to code pages on the printer), and the actual code page data (which map UTF-8 to the code page characters). So by storing our own list, we have a single source of truth for how each code page can be generated, keeping the profiles as simple as possible.

Also, based on what I've typed out so far, it does not appear safe to assume that `"".encode()` will be sufficient on its own (there is no lowest-common-denominator for printer code page support). See also [encoding.yml](https://github.com/mike42/escpos-snippets/blob/553eb569af50d552ad04f994144eb1ffd5f3c3c6/capability-db/data/encoding.yml)

miracle2k:
> I used to see a lot of bug report traffic from Windows users who were struggling to encode. An encoding cache is now bundled with the escpos-php library, so iconv is no longer a run-time dependency. 

That's clever.

> So by storing our own list, we have a single source of truth for how each code page can be generated, keeping the profiles as simple as possible.

To be sure, I do agree with your suggestion of moving the code pages to a distinct data structure (albeit from what you described, I could see a situation where one printer implements a codepage differently from others; we can solve this when it occurs).

> Also, based on what I've typed out so far, it does not appear safe to assume that "".encode() will be sufficient on its own (there is no lowest-common-denominator for printer code page support).

In what way? Because some of the encodings might differ in the printer implementation?

I don't know. I guess it doesn't hurt to have a file with the `CP857: python_encode: cp857` mappings. But I'm not sure if I would use them for implementing the encodings in Python.

If one of the Python encodings is wrong, and doesn't match what the printer implements, then of course we should handle this. But this could easily be a short "blacklist" array  put directly into the code. 

Even better of course would be one of those "data" char matrixes, so the encoding could be supported correctly.

But if the printer info told me that the printer understands "CP868", I would simply want to check ``str.encode()`` as the authoritative source if this encoding is supported or not. Maybe someone installed a module that adds extra encodings. The yaml file doesn't seen to offer any advantage.

mike42:
Continuing to workshop this, I think we're close to something final. After more check-overs, looks like python string encode plus a few typed-out pages should have good language coverage. Should be easy to write a script to export the code page data later on if we run into problems.

I noticed some changes were needed to the YAML file as the ideas have evolved. Can you take a look at-

- Small skeleton project. Is CC-Attribution ok? https://github.com/mike42/escpos-printer-db
- Profile spec format-in-progress that humans can hopefully understand: https://github.com/mike42/escpos-printer-db/tree/6d8726f8d3729b785ddcee236bd1ee8cb1827d6d/data
- Compiled `capabilities.yml` files in formats that both projects can hopefully use- https://github.com/mike42/escpos-printer-db/tree/6d8726f8d3729b785ddcee236bd1ee8cb1827d6d/dist

Main changes from your latest format is-

- Inheritance is getting stripped out in the compiled file, so the drivers don't have to deal with it at all
- everything gets an ID (eg. fonts have 0, 1 2 in the background, 'kanji' is going to be hard to send a command for, as map ordering will get lost)
- no subsets of code-page support, just overriding the whole map if it's different

Still have a lot of profiles to make/test/update, but hopefully we can move forward with this pull request while that happens in the background.

patkan:
Sorry for joining the party late :-)
I really like where this is heading.

https://github.com/mike42/escpos-printer-db looks good. Later on we could automatically create the collated yml via travis or via a post-commit-hook or something like this.
The license is ok for me, but I think CC is not that good for code-ish things. Maybe MIT is better? However, I am open there.
Just a suggestion, it is also ok like this: We could move the project to the python-escpos organization or create a new organization.

From my point of view using the str.encode()-function should suffice in python for the most usecases.
So I think the version that is implemented in the repo by @mike42 is OK. 

I lost the overview a bit. Which points are still unclear? Are there still any?

miracle2k:
I like everything Mike did there. Thanks a lot for the hard work.

CC is fine with me; this is more a content-project than code, which I why I think Mike suggested it. I assume that CC should be compatible with the code licenses too? (As an aside, python-escpos is GPL licensed; I'm not a fan of that).

Personally, I still think the `python_encode` and `iconv` mapping is unnecessary, at least in those cases where the the two strings match anyway, but it seems Mike believes them to be useful, so I'm not going to argue that any more; they don't hurt.

patkan:
OK, I can follow that argumentation. In this sense CC is ok for me too. As I understand it, CC should be compatible with GPL. As we speak of that: I am not a huge fan of GPL too. We should probably discuss that in an issue, but I am basically open for a change. (However, we have to ask all contributors)

Concerning the mapping, I also don't think that at least python-escpos will need it. I would say we should discuss this after this is implemented on both sides. Then we will see whether it is necessary.

Thank you to you both for the good work :)

miracle2k:
I am closing this, as there is now an implementation of MagicEncode that uses the capabilities database, in #170.

PR #161, opened 2016-08-24
Support multiple tab positions

miracle2k:
### Contributor checklist

  - e.g. Epson T88IIIP

---
### Description

I had some trouble understanding how tabs are supposed to work. I'm a ESCPOS newb; that said, here are some changes that make sense to me:
- My T88IIIP that I am working with certainly supports multiple tab positions.
- They are not limited to 16 either.
- It seems unnecessary that for every control(), we generate a command to initialize the tab positions. Now the tab position needs to be explicitly initialized by calling control() with the pos argument. This is a backwards-incompatibility issue.

There may still be some Py3/Py2 issues here, but I wanted to get an opinion first.

### DIFF
--- a/src/escpos/escpos.py
+++ b/src/escpos/escpos.py
@@ -617,7 +617,7 @@ class Escpos(object):
         else:  # DEFAULT: DOES NOTHING
             pass
 
-    def control(self, ctl, pos=4):
+    def control(self, ctl, pos=None):
         """ Feed control sequences
 
         :param ctl: string for the following control sequences:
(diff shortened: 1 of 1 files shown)


patkan:
In my opinion the 'forced side effects' of the methods should be as minimal as possible. Thus, removing the mandatory initialization of tabs is a good one. The best option would probably removing all the other side-effects (e.g. those in set()), cleaning up the API and then releasing a version 3.
Since I did not touch that method yet it is quite plausible that it is not completely following ESCPOS. I will have a look at the command reference as soon as I find time.

patkan:
Does this really work on your printer?
As I understand the ESC/POS-definition you have to add a NUL-byte to the line for the printer to realize it is ending. Look here: https://www.epson-biz.com/modules/ref_escpos/index.php?content_id=53 (requires probably a login)
There are also some other things that we might want to implement. (e.g. tabs only every 8th character, maximum of 32 tabs and so on)

patkan:
closed in favor to #238

PR #164, opened 2016-08-26
Constants

miracle2k:

These aren't currently used, but it's good to have them. They are taken from:

https://github.com/shantanubhadoria/python-printer-escpos


patkan:
This can be merged now, can't it?

PR #165, opened 2016-08-26
Capabilities profiles

miracle2k:
### Contributor checklist

 * e.g. Epson TM-T88II

----------

This is similar to the profiles in php-escpos, except I want to collect information about the columns on each device, too. I think YAML is a good format for this, as it's all static information, and we can share it with the PHP project: https://github.com/mike42/escpos-php/issues/160

It allows us to automatically select the right bar code function, or use the correct device width for text wrapping.

Missing:

- [x] Tests
- [x] Include the YAML file in the build
- [x] Make the barcode function use the profile.


patkan:
I really like that idea. Ideally we would then move the YAML-files into an own repository. This way both projects could simply include the capabilities as a subrepository.

miracle2k:
The current state is as follows:

- Uses the current JSON build of mike42/escpos-printer-db.
- Uses this information in ``block_text`` (column count), validating font name and barcode type.
- It has new tests for the barcode function.
- It switches to pytest as the test runner, though nose assert functions are still used for now.

I think this is ready to be merged. Once it is, I would start working on making MagicEncode use the encoding info from the profiles.

patkan:
Hi,
looks good so far, just some points:

* [ ] I would be happy if you could address the minor issues that quantifiedcode finds
* [ ] If I understand right, you have switched from branch-coverage to line-coverage. Is this intentional? If yes, why?
* [ ] I would prefer the capabilites.json as a git-submodule. This can however be done later.

miracle2k:
If you can do the submodule stuff, I'd be happy, since I don't really know how they work.

Another reason why I decided to copy the file is that i compressed it by removing whitespace. This may or may not be sensible/a good idea/necessary.

patkan:
I am not sure, whether the compression is necessary, but it probably does not disturb.
How did you compress the files? (With which tool?)

patkan:
I have thought a bit more about that. I don't think we need a minified json. Unminified it would still be understandable for someone reading the source distribution.
Minifying does only safe space. The json is however stored on the local filesystem and not transferred over the Internet. So this is not really necessary.

PR #181, opened 2016-10-04
Swap default image fragment height to 960

mike42:
### Contributor checklist

 * e.g. Epson TM-T88II

----------

The image formats in ESC/POS handle (variously) rows of 1, 8 or 24 rows of pixels at a time. The default image fragment height is currently 1024. Some image formats used in ESC/POS are not capable of representing this size without padding the image (eg, high density column-format bit images).

This pull request changes the default fragment height to 960, which is a multiple of 24, so that vertical white-space is not sent to the printer for long images.


patkan:
Sounds reasonable, thank you for the fix.

PR #204, opened 2017-02-03
Handle cases when fullCut or partCut not available

katyukha:
### Contributor checklist

 * e.g. NETUM NT-5890K

----------

### Description

Related [PR](https://github.com/receipt-print-hq/escpos-printer-db/pull/25) introduces new features:
- paperPartCut
- paperFullCut

This PR handles them with minimal code changes.

### DIFF
--- a/src/escpos/escpos.py
+++ b/src/escpos/escpos.py
@@ -564,7 +564,7 @@ class Escpos(object):
 
         self._raw(LINESPACING_FUNCS[divisor] + six.int2byte(spacing))
 
-    def cut(self, mode=''):
+    def cut(self, mode='FULL'):
         """ Cut paper.
 
         Without any arguments the paper will be cut completely. With 'mode=PART' a partial cut will
@@ -573,17 +573,25 @@ class Escpos(object):
 
         .. todo:: Check this function on TM-T88II.
 
-        :param mode: set to 'PART' for a partial cut
+        :param mode: set to 'PART' for a partial cut. default: 'FULL'
+        :raises ValueError: if mode not in ('FULL', 'PART')
         """
-        # Fix the size between last line and cut
-        # TODO: handle this with a line feed
-        self._raw(b"\n\n\n\n\n\n")
+        self.print_and_feed(6)
 
-        if mode.upper() == "PART" and self.profile.supports('paperPartCut'):
-            self._raw(PAPER_PART_CUT)
-        elif mode.upper() != "PART" and self.profile.supports('paperFullCut'):
-            # DEFAULT MODE: FULL CUT
-            self._raw(PAPER_FULL_CUT)
+        mode = mode.upper()
+        if mode not in ('FULL', 'PART'):
+            raise ValueError("Mode must be one of ('FULL', 'PART')")
+
+        if mode == "PART":
+            if self.profile.supports('paperPartCut'):
+                self._raw(PAPER_PART_CUT)
+            elif self.profile.supports('paperFullCut'):
+                self._raw(PAPER_FULL_CUT)
+        elif mode == "FULL":
+            if self.profile.supports('paperFullCut'):
+                self._raw(PAPER_FULL_CUT)
+            elif self.profile.supports('paperPartCut'):
+                self._raw(PAPER_PART_CUT)
 
     def cashdraw(self, pin):
         """ Send pulse to kick the cash drawer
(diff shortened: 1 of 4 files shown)


src/escpos/escpos.py
```diff
@@ -578,9 +578,11 @@ def cut(self, mode=''):
         # Fix the size between last line and cut
         # TODO: handle this with a line feed
         self._raw(b"\n\n\n\n\n\n")
-        if mode.upper() == "PART":
+
+        if mode.upper() == "PART" and self.profile.supports('paperPartCut'):
             self._raw(PAPER_PART_CUT)
-        else:  # DEFAULT MODE: FULL CUT
+        elif mode.upper() != "PART" and self.profile.supports('paperFullCut'):
```
patkan: You probably mean "FULL".

patkan: Moreover the full cut should also happen in the case that `mode` is not supplied.

For the case that `paperFullCut` is not available but `paperPartCut` is, defaulting to a partial cut would be good.

katyukha:
Refactored `cut` method. Also it now uses new `print_and_feed` method, which implements `ESC d n` command ([doc](http://content.epson.de/fileadmin/content/files/RSD/downloads/escpos.pdf) and [doc2](http://www.starmicronics.com/support/mannualfolder/escpos_cm_en.pdf) used for it).
Is it right way, or it is better to use old code (using `\n\n\n\n\n\n` str instead)?

src/escpos/escpos.py
katyukha: Seems to be implemented, to use one available method. But i am not sure if it is right to use FullCut if requested PartCut. May be just ignore it, if PartCut not available on printer?

katyukha:
Related PR [receipt-print-hq/escpos-printer-db/pull/25](https://github.com/receipt-print-hq/escpos-printer-db/pull/25) is merged. code updated.

Is it Ok?

patkan:
Hi @katyukha thank you!
Looks good to me. I am not sure about some names, but I think it is best to nail those down when we have to touch the complete API for the v3-release.

Just one last idea: How about issuing a `LF` instead of `ESC d n` in the case of 1 line? I don't know the level of support of this command and this way we would ensure that printing at least one line works.

katyukha:
This commit fixes bug in `control` method, which makes using `control('LF')` have not result.

Also, it seems that `cut` command here is implemented incorrectly due to [ESC/POS doc](http://content.epson.de/fileadmin/content/files/RSD/downloads/escpos.pdf)

Here is following command code is used:

```python
# Paper Cutter
_CUT_PAPER = lambda m: GS + b'V' + m
PAPER_FULL_CUT = _CUT_PAPER(b'\x00')  # Full cut paper
PAPER_PART_CUT = _CUT_PAPER(b'\x01')  # Partial cut paper
```

But, as documentation says, there are two possible functions signatures for `cut`:

Function A: ```GS V m```
Function B: ```GS V m n```

Where ```m``` is:
- (Function A) m = 0, 1, 48, 49
- (Function B m = 65, 66

And ```n``` for function B is number of lines to feed.

So for **Function A**, possible commands:

```python
_CUT_PAPER = lambda m: GS + b'V' + m
PAPER_FULL_CUT = _CUT_PAPER(b'\x00\x30')  # Full cut paper
PAPER_PART_CUT = _CUT_PAPER(b'\x01\0x31')  # Partial cut paper
```

And for **Function B**, possible commands:

```python
_CUT_PAPER = lambda m, n=1: GS + b'V' + m + six.int2byte(n)
PAPER_FULL_CUT = _CUT_PAPER(b'\x41', n)  # Full cut paper
PAPER_PART_CUT = _CUT_PAPER(b'\x42', n)  # Partial cut paper
```

Also, may be it is a good idea to add number of lines to feed as parametr to cut?

As for names? You mean rename method `print_and_feed` for example to `feed`?

patkan:
Hi @katyukha, sorry for taking so long. I wanted to merge your contribution, but I've got problems with your fix to `cut`. For my TM-T88II this does not work. It will print a 0 or 1 when cutting. Are you sure with the `\x31`. Where does this come from? You cite the reference but I can't find a mention of said sequence. (for cut it is e.g. just `GS V 0` according to the reference for mode A)

If this is needed for your printer, then we will have to add it as a capability. In this case I would like to do this in a different PR to get this one merged.

katyukha:
Hi @patkan ! Yes, last commit seems to be incorrect( So i will revert it back.

patkan:
Merged, thank you very much!

Issue #96, opened 2016-03-06
improve set-method


patkan:
Improve the set-method just like in e1960c896527f3c6b201ba46045bb769bb5293f0.
(This would then change the behaviour of the class, but also makes it better testable.)

MicroJoe:
See PR #216 for a WIP of this issue.

patkan:
Sorry, this has not been linked, but is also relevant #176. (And also sorry for the very short description of the issue)

MicroJoe:
Should we close this?

I know someone wanted a method like set but without reseting everything for efficiency, but this need should be expressed clearly in another issue IMO.

patkan:
You are right, this is just undefined. I'm gonna create a new issue.

PR #216, opened 2017-05-14
Started refactor of the set method

MicroJoe:
### Contributor checklist

 * Tested refactor using hand-created tests for before-and-after regression avoidance

----------

### Description

This is a work in progress addressing the #96 issue.

I am starting it here in order to see if people are interested in this before further development. This allows the set method to be tested (I will write the tests before modifying a feature, then refactor, then rerun the test).

### DIFF
--- a/src/escpos/constants.py
+++ b/src/escpos/constants.py
@@ -71,50 +71,89 @@ SHEET_ROLL_MODE = ESC + b'\x63\x30\x01'  # paper roll
 # Text format
 # TODO: Acquire the "ESC/POS Application Programming Guide for Paper Roll
 #       Printers" and tidy up this stuff too.
-TXT_FLIP_ON    = ESC + b'\x7b\x01'
-TXT_FLIP_OFF   = ESC + b'\x7b\x00'
-TXT_SMOOTH_ON  = GS + b'\x62\x01'
-TXT_SMOOTH_OFF = GS + b'\x62\x00'
 TXT_SIZE       = GS + b'!'
-TXT_WIDTH      = {1: 0x00,
-                  2: 0x10,
-                  3: 0x20,
-                  4: 0x30,
-                  5: 0x40,
-                  6: 0x50,
-                  7: 0x60,
-                  8: 0x70}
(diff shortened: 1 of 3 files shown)


MicroJoe:
Note that this PR will break the current set API, but this seems to be ok since final 0.3 is not released yet and still open to API breaks.

MicroJoe:
I have finished the refactoring, and added tests than could run on the previous set method and that passes on the new set method too.

I have edited the PR to mark it as ready to merge since I have finalised the PR. Feedbacks very welcome.

MicroJoe:
Requesting a review for this PR (ping @patkan?)

I refactored using the dict structure mentionned in the issue, but could not factorize out the write in a for loop because of handling the special cases that would make the code unlegible.

MicroJoe:
Ok, I think I get it now. What I have wrote will be the coming `layout` method, and we will have to create a `set` method that is a bit smarter than that. This is a startpoint for future PRs I guess.

I did not test this on my printer, is there a script that tries everything on the printer so we can detect bugs? Or should I run some manual tests using the `set` method?

I am also thinking of some refactor around the `size` parameter to be splitted into two boolean properties: `double_width=False` and `double_height=False` by default. Doing so will avoid string parameters and use booleans that are easier to reason about.

patkan:
Unfortunately there is no test script. A good place to keep something like this would be the examples folder.
Since this is more of a refactoring than a real change of "escpos-code" testing on a real printer is not that essential, I just wanted to try it out.

Creating more booleans is a good one. This also makes auto-completition and similar stuff easier.

MicroJoe:
Here is the booleans commit. I think I will stop here for this PR and start to work on other issues.

patkan:
Sorry, I've been busy.
It got around to trying it out and it looks good. Thank you!

MicroJoe:
> It got around to trying it out and it looks good. Thank you!

Maybe you should add your test code to the examples directory for other people to test their printers against text formatting.

Thanks for the merge!

patkan:
I would like to, but I tested by playing around in a ipython shell.
I will try to think of writing a script next time.

PR #219, opened 2017-05-15
Preliminary support of pos 'line display' printing through Epson printer

patkan:
This pulls in the changes by @Ousret.
This is useful for setups that have a display in the chain before the printer. It selects the display, resets it, sends text and then again selects the printer.


patkan:
closing in favor of #224

Issue #218, opened 2017-05-14
Barcode software printing


patkan:
Some printers like in #58 do not support barcode printing or only support a subset of barcodes (as the old Epson printers).

A possible enhancement would be finding a proper library for barcode-printing. ([python-barcode](https://bitbucket.org/whitie/python-barcode)?) With this barcode could be rendered in software and then printed as images.
Ideally a new capability in https://github.com/receipt-print-hq/escpos-printer-db would be used to determine whether software-rendering is needed. (See the QR-function as an example)

icarolongo:
Anyone tried to print QR Code on POS-5890?
It uses python qrcode correctly?

mike42:
This thread is about barcodes, not qrcodes. I tested the software qrcode
feature on a ZJ5890K (or similar) when I implemented it, please file a new
issue if you think you've found a bug or opportunity for improvement!

MicroJoe:
Ok so #222 was merged.

I looked into the hardware barcode method for a possible merge with the software defined method. The hardware method has a lot of hardware-specific arguments such as font type, A or B barcode mode, etc..

So, the software render of barcodes is available but is not in the same method as the hardware barcode. Merging the two methods within one would using a `use_software` boolean argument is possible but needs some refactoring of the hardware method interface.

Maybe we can close this issue and create a new one if we want the two methods to be merged into one with a work on the method arguments.

patkan:
I think the real task is splitting the library into low-level Escpos functions (or maybe also those emulating them) and highlevel "layout" functions.
Relevant issues are #176 and #96.

We would have to start with a concept and then implement it. But yeah, considering this issue I think we are done.

PR #222, opened 2017-05-16
First implementation of software barcode

MicroJoe:
### Contributor checklist

 * ZJ5890

----------

### Description

Actually the hardware barcode implementation is very specific and not generic enough for just adding a `soft_render=True` argument to it. This is a first work that can be improved with other commits, maybe for
merging this method in the `barcode` method after some cleanup.

The width, height and text_distance were set using empiric print-and-retry tests so that the generated barcode looks nice to the eye (and to the eye of an Android scanner tool.

!WARNING! Printing a barcode that is too large in width will result in the printer to go crazy trying to print an image that is too large for it. This may be fixed by raising an exception in the `image` method.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -115,7 +115,8 @@ setup(
         'pyyaml',
         'argparse',
         'argcomplete',
-        'future'
+        'future',
+        'pyBarcode==0.8b1'
     ],
     setup_requires=[
         'setuptools_scm',
--- a/setup.py
+++ b/setup.py
@@ -116,7 +116,7 @@ setup(
         'argparse',
         'argcomplete',
         'future',
-        'pyBarcode==0.8b1'
+        'viivakoodi>=0.8'
     ],
     setup_requires=[
         'setuptools_scm',
--- a/src/escpos/escpos.py
+++ b/src/escpos/escpos.py
@@ -19,6 +19,9 @@ import qrcode
 import textwrap
 import six
 
+import barcode
+from barcode.writer import ImageWriter
+
 from .constants import ESC, GS, NUL, QR_ECLEVEL_L, QR_ECLEVEL_M, QR_ECLEVEL_H, QR_ECLEVEL_Q
 from .constants import QR_MODEL_1, QR_MODEL_2, QR_MICRO, BARCODE_TYPES, BARCODE_HEIGHT, BARCODE_WIDTH
 from .constants import TXT_ALIGN_CT, TXT_ALIGN_LT, TXT_ALIGN_RT, BARCODE_FONT_A, BARCODE_FONT_B
@@ -395,6 +398,26 @@ class Escpos(object):
 
         if function_type.upper() == "A":
             self._raw(NUL)
+
+    def soft_barcode(self, barcode_type, data, module_height=5, module_width=0.2, text_distance=1):
+        image_writer = ImageWriter()
+
+        if barcode_type not in barcode.PROVIDED_BARCODES:
+            raise BarcodeTypeError(
+                'Barcode type {} not supported by software barcode renderer'
+                .format(barcode_type))
+
+        barcode_class = barcode.get_barcode_class(barcode_type)
+        my_code = barcode_class(data, writer=image_writer)
+
+        my_code.write("/dev/null", {
+            'module_height': module_height,
+            'module_width': module_width,
+            'text_distance': text_distance
+        })
+
+        image = my_code.writer._image
+        self.image(image, impl='bitImageColumn')
 
     def text(self, txt):
         """ Print alpha-numeric text
--- a/src/escpos/escpos.py
+++ b/src/escpos/escpos.py
@@ -399,14 +399,18 @@ class Escpos(object):
         if function_type.upper() == "A":
             self._raw(NUL)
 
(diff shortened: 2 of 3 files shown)


src/escpos/escpos.py
```diff
@@ -396,6 +399,26 @@ def barcode(self, code, bc, height=64, width=3, pos="BELOW", font="A",
...
+        barcode_class = barcode.get_barcode_class(barcode_type)
+        my_code = barcode_class(data, writer=image_writer)
+
+        my_code.write("/dev/null", {
+            'module_height': module_height,
+            'module_width': module_width,
+            'text_distance': text_distance
+        })
+
+        image = my_code.writer._image
```
patkan: Maybe we could patch the writer to expose the image via an public method (instead of this private member). Otherwise, for now this is OK for me.

src/escpos/escpos.py
```diff
@@ -396,6 +399,26 @@ def barcode(self, code, bc, height=64, width=3, pos="BELOW", font="A",
...
+        self.image(image, impl='bitImageColumn')
```
patkan: Forcing an implementation in this method makes the code unusable for certain printers.
Unfortunately the capabilities-implementation does not support (yet) automatic selection of the image implementation.
Maybe let the user choose the image-implementation when calling the soft_barcode-method?

setup.py
```diff
@@ -115,7 +115,8 @@ def run_tests(self):
         'pyyaml',
         'argparse',
         'argcomplete',
-        'future'
+        'future',
+        'pyBarcode==0.8b1'
```
MicroJoe: The 0.8 version of pyBarcode introduced code128 support and I really want this support into our implementation. [Latest PyPI version is 0.8b1](https://pypi.python.org/pypi?%3Aaction=search&term=pyBarcode&submit=search) but if I put the `>=0.8` requirement the beta version is not retrieved:

```
pip install 'pyBarcode>=0.8'
Collecting pyBarcode>=0.8
  Could not find a version that satisfies the requirement pyBarcode>=0.8 (from versions: 0.5.1, 0.6, 0.7, 0.8b1)
No matching distribution found for pyBarcode>=0.8
```

The real solution is to make pyBarcode release a v0.8 tag and push it on PyPI but the [pyBarcode repository seems quite inactive right now](https://bitbucket.org/whitie/python-barcode). I emailed the maintainer but had no response so far ; maybe we should fork it and rename it?

src/escpos/escpos.py
```diff
@@ -396,6 +399,26 @@ def barcode(self, code, bc, height=64, width=3, pos="BELOW", font="A",
         if function_type.upper() == "A":
             self._raw(NUL)
 
+    def soft_barcode(self, barcode_type, data, module_height=5, module_width=0.2, text_distance=1):
```
MicroJoe: Yes, the goal is to merge this method with `barcode` in the long run.

src/escpos/escpos.py
MicroJoe: We should patch pyBarcode upstream in order to add a property to the writer that returns the Pillow image. Since upstream seems dead maybe this is an opportunity to fork or find forks of pyBarcode that are more active.

src/escpos/escpos.py
MicroJoe: Added, but I do not garantee the results using anything else than this impl.

setup.py
patkan: OK, then we can keep it like this.

After reading [pyBarcode#16](https://bitbucket.org/whitie/python-barcode/issues/16/pypi-08-release-request) it looks like this project is discontinued. However it seems to be continued in [this project](https://github.com/kxepal/viivakoodi).
So maybe we can switch the dependency and then improve this project.

patkan: Another option might be [steenzout](https://github.com/steenzout/python-barcode). It seems to have a bit more traction. @MicroJoe, could you maybe check both projects and decide which one is better?

MicroJoe: I fired some issues on viivakoodi but then after saw that the steenzout fork was way more maintained. I think this is the best option, I will try to change dependencies to this one.

MicroJoe: Well after looking a little bit more the steenzout fork does not seems very community-friendly as the author of the fork is using his own libraries as requirements. I am more happy to contribute to the viivakoodi project that has no ambiguity about the package name, and seems less one-user-centric.

patkan: I am not sure about the steenzout-fork being community-unfriendly, but the only benefit of this fork seem to be some unit tests and a CLI. Unit tests can also be added and we don't need a CLI.

Because of the lacking ambiguity I would also tend to use viivakoodi. I think it is best to wait if the project is still alive and then you can change the dependecy here.

MicroJoe:
Related to #218

setup.py
MicroJoe: Created new issue on viivakoodi: https://github.com/kxepal/viivakoodi/issues/12

MicroJoe:
I have replaced pyBarcode with viivakoodi. May we change the fork when we will need it, for now it is going to do the job and we can focus on the compatibility with the already existing barcode function once this is merged.

patkan:
Thanky you!

PR #235, opened 2017-07-18
 Ensure QR codes have a border large enough

### LINKED ISSUE #29: The qr() method must whitespace-pad above and below the QR code
```
The white border around a QR code is a non-optional part of the registration 
marking system used by scanners to recognize the code.

In order to get the libdecodeqr-simpletest example program to recognize the QR 
code printed by the example code at 
https://code.google.com/p/python-escpos/wiki/Usage and decode it, I had to:

1. Scan it
2. Open it in GIMP
3. Use the eraser tool to replace part of the logo.gif contents above and 
barcode below with white.

While I understand taking the full 13 digits of an EAN13 barcode as input since 
barcode readers and ISBN-13 databases treat the checksum digit as part of the 
data, there is no similar reason to let users generate broken QR codes.
```

Original issue reported on code.google.com by `stephan....@gmail.com` on 20 May 2014 at 1:59

MicroJoe:
(The QR code spec requires a border at least 4*box_size thick but we can't
 just set border=16 because that results in a QR code more than 255px tall
 and I'm not yet ready to use fullimage() as a backend for it)

This fix was originally commited by Stephan Sokolow on 2014-05-22

### Contributor checklist

 * ZJ-5890

----------

### Description

Fixes #29.

The centering does not work so it was removed from original commit 2a8f852.


patkan:
Thank you :-)
Maybe you already have seen it: this breaks the test comparing the output string of the function. So this has to be fixed before merging.

Apart from that, why do we need 1 newline before and 2 after the QR code?

MicroJoe:
Here are some commits that may fix the tests. I have introduced newlines in the test so that we can clearly see the newlines appart from the QR code data.

The two line feeds were needed after the QR because the first one seems to end the image print but does not add a newline. Here is a photo of the result:

escpos-qr

You can see that the spacing seems even.

PR #236, opened 2017-07-18
Optional feed for cut (closes #213)

### LINKED ISSUE #213: Cut paper without feeding it
Hello alltogether,

currently i'm facing a problem with the .cut()-function.
I would like to achieve a cut of the paper without feeding, so i do not have approx. 4 lines at the top of each next sheet.

Inside of  _python-escpos/src/escpos/constants.py_ between lines 54-57 i can see the esc sequence for full and partial cut but i do not find any code for cutting without feeding.

So i tried to send raw data (GS V 66 0) to the printer, but i got _AttributeError: 'Usb' object has no attribute 'raw'_.

(How) is it possible to make a cut at the end one "sheet" without having blank lines on the top of the next sheet?

Due to the fact that i'm not a (good) programmer i apologize for this maybe stupid question...

Best regards!

I have:

### Device info

 **Controler:** Epson BA-T500II USB
 **Printer:** M-T522AF II

 **python-escpos version:** 2.2.0

 **python version:** 3.4.2

 **operating system:** Raspbian GNU/Linux 8 (jessie)

MicroJoe:
### Contributor checklist

 * e.g. Epson TM-T88II

----------

### Description

Optional feed. Fixes #213.

### DIFF
--- a/src/escpos/escpos.py
+++ b/src/escpos/escpos.py
@@ -563,7 +563,7 @@ class Escpos(object):
 
         self._raw(LINESPACING_FUNCS[divisor] + six.int2byte(spacing))
 
-    def cut(self, mode='FULL'):
+    def cut(self, mode='FULL', feed=True):
         """ Cut paper.
 
         Without any arguments the paper will be cut completely. With 'mode=PART' a partial cut will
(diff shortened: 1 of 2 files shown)


patkan:
Before I merge this, I would be interested, why the code here differs from the code that I proposed in the corresponding issue. If you know it you can explain it to me, otherwise I will just read the spec on cutting and feed as soon as I get around to it.

MicroJoe:
I did this patch expecting that not calling `print_and_feed`should be enough but it was not the case. I have added a test so I can now guarantee that we have the correct output that you mentioned in your issue.

Since I do not have a printer that support cut I cannot test this, but the binary output is now tested so it should be ok.

patkan:
Thank you! :-)

PR #237, opened 2017-07-22
Added the DLE EOT querying command.

csoft2k:
Added a function to check whether the printer is online or not, as well
as a reading method for USB printers.

### Contributor checklist

- [x] I have tested my contribution on these devices: Epson TM-T20II

----------

### Description
As of now this module lacked a reading capability to check the status of the printers.
I thought it could be great to be able to check if the printer is online or not.
This can be achieved by invoking the printer._isOnline() method, which returns a Boolean - if True, the printer is (obviously...) online.
I've only been able to develop this in the Usb class, since I lack any other interfaces for my printer.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -10,7 +10,6 @@ Dean Rispin
 Dmytro Katyukha
 Hark
 Joel Lehtonen
-K
 Kristi
 ldos
 Manuel F Martinez
--- a/src/escpos/escpos.py
+++ b/src/escpos/escpos.py
@@ -18,6 +18,7 @@ from __future__ import unicode_literals
 import qrcode
 import textwrap
 import six
+import time
 
 import barcode
 from barcode.writer import ImageWriter
@@ -34,6 +35,7 @@ from .constants import CD_KICK_DEC_SEQUENCE, CD_KICK_5, CD_KICK_2, PAPER_FULL_CU
 from .constants import HW_RESET, HW_SELECT, HW_INIT
 from .constants import CTL_VT, CTL_HT, CTL_CR, CTL_FF, CTL_LF, CTL_SET_HT, PANEL_BUTTON_OFF, PANEL_BUTTON_ON
 from .constants import TXT_STYLE
+from .constants import RT_STATUS_ONLINE, RT_STATUS_OFFLINECAUSE, RT_MASK_ONLINE
 
 from .exceptions import BarcodeTypeError, BarcodeSizeError, TabPosError
 from .exceptions import CashDrawerError, SetVariableError, BarcodeCodeError
@@ -733,6 +735,18 @@ class Escpos(object):
         else:
             self._raw(PANEL_BUTTON_OFF)
 
+    def queryStatus(self):
+        self._raw(RT_STATUS_ONLINE)
+        time.sleep(1)
+        status = self._read();
+        if (len(status) > 0):
+            return status;
+        else:
+            return [8];
+
+    def _isOnline(self):
+        status = self.queryStatus()[0];
+        return ((status & RT_MASK_ONLINE) == 0);
 
 class EscposIO(object):
     """ESC/POS Printer IO object
--- a/src/escpos/escpos.py
+++ b/src/escpos/escpos.py
@@ -79,6 +79,9 @@ class Escpos(object):
         """
         pass
 
+    def _read(self, msg):
+        raise NotImplementedError(); 
+
     def image(self, img_source, high_density_vertical=True, high_density_horizontal=True, impl="bitImageRaster",
               fragment_height=960):
         """ Print an image
@@ -744,7 +747,7 @@ class Escpos(object):
         else:
             return [8];
 
-    def _isOnline(self):
+    def isOnline(self):
         status = self.queryStatus()[0];
         return ((status & RT_MASK_ONLINE) == 0);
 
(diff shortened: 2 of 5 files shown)


src/escpos/printer.py
```diff
@@ -84,6 +84,10 @@ def _raw(self, msg):
         """
         self.device.write(self.out_ep, msg, self.timeout)
 
+    def _read(self):
```
patkan: You should add a dummy implementation to the base class that throws a NotImplementedError.

src/escpos/escpos.py
```diff
@@ -733,6 +735,18 @@ def panel_buttons(self, enable=True):
...
+    def queryStatus(self):
+        self._raw(RT_STATUS_ONLINE)
+        time.sleep(1)
+        status = self._read();
+        if (len(status) > 0):
+            return status;
+        else:
+            return [8];
+
+    def _isOnline(self):
```
patkan: Why do you use a underscore here for the name. Isn't this intended to be called?

src/escpos/printer.py
patkan: This way the other implementations will still function without crashing...

src/escpos/escpos.py
csoft2k: This has been my first-time Python coding - wasn't sure about the convention.
Thanks!

src/escpos/printer.py
csoft2k: Done, thanks for finding out this.

src/escpos/escpos.py
```diff
@@ -733,6 +738,18 @@ def panel_buttons(self, enable=True):
...
+        time.sleep(1)
+        status = self._read();
+        if (len(status) > 0):
+            return status;
+        else:
+            return [8];
+
+    def isOnline(self):
+        status = self.queryStatus()[0];
+        return ((status & RT_MASK_ONLINE) == 0);
```
belono: Semicolons are not needed at the end of a sentence.
Also, syntax could be more pythonic changing '&' with 'and'.

``
return ((status and RT_MASK_ONLINE) == 0)
``

src/escpos/escpos.py
```diff
@@ -733,6 +738,18 @@ def panel_buttons(self, enable=True):
...
 
+    def queryStatus(self):
+        self._raw(RT_STATUS_ONLINE)
+        time.sleep(1)
+        status = self._read();
+        if (len(status) > 0):
+            return status;
+        else:
+            return [8];
+
```
belono: The if-else block could be avoided using:

``
return status or [8]
``

src/escpos/escpos.py
belono: Method names should be 'is_online' and 'query_status'.

patkan:
I second what @belono wrote and would also ask you to check the [flake8-errors](https://travis-ci.org/python-escpos/python-escpos/jobs/256333711) in the corresponding travis check.

AUTHORS
```diff
@@ -3,12 +3,14 @@ Asuki Kono
...
 Christoph Heuel
 Cody (Quantified Code Bot)
+csoft2k
 Curtis // mashedkeyboard
 Davis Goglin
 Dean Rispin
 Dmytro Katyukha
 Hark
 Joel Lehtonen
+K
```
patkan: You can use the .mailmap file to cluster your different usernames together

src/escpos/escpos.py
csoft2k: All changes have been pushed but the & -> and one. If I'm not wrong, the operator must be '&' since a binary 'and' is needed.

src/escpos/escpos.py
csoft2k: Done. Thanks - didn't know that syntax.

AUTHORS
csoft2k: I feel I might need some help to achieve this. Should I upload that file or thee is a command to perform that clustering?

patkan: This is a [feature of git-shortlog](https://git-scm.com/docs/git-shortlog#_mapping_authors). You just have to add your name-mappings to the .mailmap-file in the root directory of this project.

So you could just add the line `csoft2k <csoft2k@hotmail.com>`, which will replace the name "K" in the one commit that you made, when creating shortlogs and similar stuff. ;-)
Otherwise, this is something that I can also fix, if you want.

src/escpos/escpos.py
patkan: Considering the &: you are totally right ;-)

belono: Oops, sorry, you're right, the operator must be '&'. I'm not familiar with bitwise operators.

Is this the only way to return the status boolean?
Maybe it could work ``return bool(status is [8])`` or similar?

patkan: an option could be `return not (status&RT_MASK_ONLINE)`

csoft2k: Well, keep in mind the printer returns some bits fixed to 1. The normal online status is:
00010110 (22)
But the only bit that marks the printer is online is the 4th (bit3): if 0, the printer is online, if 1, offline.

The bool(status is [8]) alternative doesn't look feasible to me, but I could be wrong.
Regarding changing the ==0 with a 'not', yes, it works. I'll push the change in a minute.

patkan:
Thank you for your contribution! I have merged it into the development branch.

csoft2k:
My pleasure! :)

belono:
Looks great @csoft2k, thanks.

PR #240, opened 2017-07-27
v3.02a

patkan:
release of a new alpha with the current changes


MicroJoe:
Nothing to add. We have nice features in this one! I can now send barcodes to my cheap printer.

PR #246, opened 2017-07-30
Add methods for simpler newlines

MicroJoe:
### Contributor checklist

 * e.g. Epson TM-T88II

----------

### Description

Tested only using pytest, but so obvious it should work for real too. The goal is to avoid to type `'\n'` everytime we have to make a new line.


patkan:
Thank you!

PR #252, opened 2017-08-17
Pickling capabilities for faster start up times.

serpulga:
- [x] I have tested my contribution on these devices: Epson TM-T20

#### Pickling capabilities 
Pickling `capabilities.json` for quicker start time. On a **RaspberryPi** it's taking 10 seconds to simply run: `import escpos.printer`.  This change creates a pickle file that will load **20x faster**. The rationale is that the `capabilities.json` file doesn't change too often. Also changed import order for PEP8.

#### Example file
Fixed outdated example file: should be font `'b'` not `'B'`

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -9,3 +9,4 @@ Renato Lorenzi <renato.lorenzi@senior.com.br>        Renato.Lorenzi <renato.lore
 Ahmed Tahri <nyuubi.10@gmail.com>                    TAHRI Ahmed <nyuubi.10@gmail.com>
 Michael Elsdörfer <michael@elsdoerfer.com>           Michael Elsdörfer <michael@elsdoerfer.info>
 csoft2k <csoft2k@hotmail.com>
+Sergio Pulgarin <sergio.pulgarin@gmail.com>
--- a/AUTHORS
+++ b/AUTHORS
@@ -22,6 +22,8 @@ Qian Linfeng
 Renato Lorenzi
 Romain Porte
 Sam Cheng
+Sergio
+Sergio Pulgarin
 Stephan Sokolow
 Thijs Triemstra
 Thomas van den Berg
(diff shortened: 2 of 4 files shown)


patkan:
Hi @serpulga thank you!
I would not have thougt that loading the capabilities takes so long. Your solution looks good to me. I will try it out as soon as I can and then merge.

serpulga:
Sweet. Thanks!

serpulga:
@patkan 
I forgot to ask: When I ran the command `./doc/generate_authors.sh` I was getting a different **AUTHORS** file than the build. I had to copy the output from travis so they matched.

patkan:
In which way was it different? Maybe it was not with both names? Did you maybe commit with another name? (You can group your names with the mailmap-feature of git-shortlog)

All in all I am thinking of removing the authors-file (and just generating it on the fly for documentation).

serpulga:
I'm not sure. The filename was the same: **AUTHORS**, but It looks like the command on my computer was sorting the names differently than on Travis, thus giving out a mismatch. In the end, both files seemed to contain the same list of authors but in a different order. In the end, I gave up and just copied whatever Travis was expecting on the file.

patkan:
Thank you for the contribution!

serpulga:
😀

thijstriemstra:
@serpulga why did you add the bare `logging.basicConfig()` call? Shouldn't this **only** be set in the program that uses this library, e.g. write it to a file? Because I now explicitly have to disable it for my application.

serpulga:
@thijstriemstra 
Hi, I'm just used to doing it that way. I never used `logging` in the context of a library. If you think it should be removed please do.

thijstriemstra:
ps. thanks for this contribution, I would've run into the same issue as I'm also using RaspberryPi.

serpulga:
Thank you for checking. I absolutely didn't want to intrude with you already working code.

thijstriemstra:
No worries! I was porting from v2.0 to 3.0 and came across this. The environment var was also a good choice, as I'm using a frozen pyqt5 app, making it trivial to place the file elsewhere and load from there.

Issue #100, opened 2016-03-08
Barcode validation


DavisGoglin:
There is a list of allowed data and barcode sizes in the [epson documentation](https://reference.epson-biz.com/modules/ref_escpos/index.php?content_id=128). We should be validating data length and content. 

A short term solution would be to set up regexes similar to [escpos-php here](https://github.com/mike42/escpos-php/blob/0365deee8591ac5e80f748dce06bf0082ea59833/src/Mike42/Escpos/Printer.php#L201), but in the future we might want to see if there is an external barcode library to validate checksums.

patkan:
So far I have found https://pypi.python.org/pypi/pyBarcode/0.7
This is, however, a library for image generation. (We could use it in order to generate barcodes on models that do not support barcode-printing)

Loafdude:
Should python-escpos venture beyond the escpos protocol? I think it's better to focus.
There are lots of libraries that can create barcodes. We should leave that up to the user.
As long as we can print, position and handle PIL objects I think we are good.

We should definitely do character set and length checking.

I have a barcode scanner handy here and most of the barcodes scanned fine. 

I think the printer is doing checksum for us?
Ill try and decode some manually and check

Loafdude:
Okay after quick review of UPC-A
escpos spec says you can supply 11 or 12 character data.

UPC-A spec requires 12 digits.
if 11 digits my printer calculates the check digit. 
if 12 digits my printer prints the supplied 12th character check digit.
this is the best of both worlds I'd say.

a very quick look at code 128 with both my smartphone and usb scanner seems to indicate the checksum is calculated

Considering all the barcodes in my test scanned I would say the printer is supposed to handle the checksum for us.

So my conclusion is if a user wants to supply their own checksum they simply pass it, otherwise it is calculated by the printer.
That said, maybe non-epson stuff doesnt work this way?
Regardless we shouldn't write escpos to a specific printer but keep it generic.
If a user's printer does not support checkdigit calculation than they will need to do it in their own code.
Perhaps you guys can check your printers to observe their behavior.

I used
    e.text("UPC-A\n")
    e.barcode('13243546576','UPC-A',function_type="A")
    e.text("UPC-A\n")
    e.barcode('132435465767','UPC-A',function_type="A")

the first should scan. (check digit printed is 2)
The second should not (check digit 7 is invalid)

regex on character set and length should be sufficient I'd say.

DavisGoglin:
[...]

PR #255, opened 2017-08-31
Feature/check barcodes

derlin:
### Contributor checklist

- [X ] I have tested my contribution on these devices: 
    * espon TM-T20II

----------

### Description

__Add a method to check the validity of a barcode code__

The validation process is the following:

1. check the length of the code,
2. check the characters and format of the code

Those checks are mainly performed using regexp and based on the ESC/POS specification (see [ESC/POS Command Specifications 4-122],
(http://www.starmicronics.com/support/mannualfolder/escpos_cm_en.pdf). This means that valid code from the barcode standard perspective might be invalid from the ESC/POS perspective and vice-versa. 

__Automatically perform the check in the `barcode()` method__

If the check fails, a `BarcodeCodeError` is raised.

__Add a parameter to disable the check in the `barcode()` method__

Add a parameter `check` (default to True). If the user want to experiment with invalid codes, he can do so but at his expenses.

### Going further

It might be nice to add support for checksums as well. 

For example, given an EAN-13 barcode code, if the code length is 12, the `barcode` method could automatically add the checksum digit. 

The problems/questions I see are:

* how about codes with a wrong checksum ?
* how about barcode standards with
     - optional checksums ?
     - variable length ?


derlin:
It seems like the commit udpating the AUTHORS list is the one failing. Is there something else to do other than running `doc/generate_authors.sh` command ?

patkan:
Hi @derlin, thanks for the contribution!

There's still some problems with the authors-check :-(
You can try to merge the current development-branch in (I just pushed a commit). Please check if this commit generates a AUTHORS-file that will pass the check. (I assume that the default settings for sort might be different)

patkan:
I just checked, if you run `./doc/generate_authors.sh` now, it should generate the file that Travis is expecting.
It might still be the case that your system handles this different. So if it does not work: could you please write which system you are using. Then I can try to research the problem.

derlin:
Ok, it worked. Thanks !

patkan:
@MicroJoe proposed in #100 to use the check_barcode-method of viivakoodi. This would be a bit contrary to this PR. What do you both (@MicroJoe and @derlin) think?

Since this is a rather contained commit, I would merge this like this. The benefit of this solution is that we can print barcodes without relying on a external library (and thus could make viivakoodi optional)

derlin:
In my opinion, it is important to ensure that a code the printer won't be able to render throws an exception. Concerning the checksum and more thorough checks, letting them to the responsibility of the user (many barcode types have only optional checks, some standards like EAN-13 have many variations...) is fine by me. Indeed, it would complexify a lot the code and can be restrictive to some users that just need barcodes for internal purposes. 

What bothers me with the current solution though is the potential difference between the actual barcode and its associated text. 

For example, calling the `barcode` function with the EAN-13 barcode `13243546576` will print the barcode `132435465762` (the ending digit `2` is the checksum computed automatically by the printer), while the printed text is still `13243546576`. 

Another example is the CODE39 standard:
```python
printer.barcode('CODE39', '*WIKIPEDIA*')
printer.barcode('CODE39', 'WIKIPEDIA')
```
Both calls print the same barcode (the start and end characters '*' are automatically added), but the text is different...

What I would propose:

* add a link to the barcode library @MicroJoe mentioned in the documentation
* automatically add checksum to the text _if it is missing_ for EAN and UPC (EAN-13 with only 12 digits -> add checksum)
* automatically add the '*' start/stop characters for CODE39 if missing
* (? make the `compute_checksum` public ?)

I am not sure about other standards, I would need to check.

patkan:
Sorry for taking so long. I got around to have a deeper look at your PR and merged it.

PR #271, opened 2017-11-25
Feature/clear content in dummy printer

braveheuel:
### Contributor checklist

 * POS-5890T

----------

### Description
When using the dummy printer as caching device, you can send the output to a physical device. But the reusing the same dummy device would result in appending the new output to already printed output. To overcome this issue, the clear function is implemented, which clears the current buffer.


patkan:
Thank you for the PR! This seems like a good idea to me :+1:

I will merge this directly after #273 is fixed.

PR #274, opened 2017-12-04
Migration to libusb1 and changes to make it compatible with Python on Windows

juanmitaboada:
… Windows

### Contributor checklist

- [ ] I have read the CONTRIBUTING.rst
 * e.g. Epson TM-T88II

----------

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,7 +3,6 @@
 import os
 import sys
 from setuptools import find_packages, setup
-from setuptools.command.test import test as test_command
 
 
 base_dir = os.path.dirname(__file__)
@@ -17,33 +16,6 @@ sys.path.insert(0, src_dir)
 def read(fname):
     """read file from same path as setup.py"""
     return open(os.path.join(os.path.dirname(__file__), fname)).read()
-
-
-class Tox(test_command):
-    """proxy class that enables tox to be run with setup.py test"""
-    user_options = [('tox-args=', 'a', "Arguments to pass to tox")]
-
-    def initialize_options(self):
-        """initialize the user-options"""
-        test_command.initialize_options(self)
-        self.tox_args = None
-
-    def finalize_options(self):
-        """finalize user-options"""
-        test_command.finalize_options(self)
-        self.test_args = []
-        self.test_suite = True
-
-    def run_tests(self):
-        """run tox and pass on user-options"""
-        # import here, cause outside the eggs aren't loaded
-        import tox
-        import shlex
-        args = self.tox_args
-        if args:
-            args = shlex.split(self.tox_args)
-        errno = tox.cmdline(args=args)
-        sys.exit(errno)
 
 
 setuptools_scm_template = """\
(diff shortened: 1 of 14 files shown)


patkan:
Hi, thank you for your contribution :-) I'm going to have a look at it later.
In order to make the AUTHORS work like you added it, you could use a mailmap line like this:
```
Juanmi Taboada // br0th3r <juanmi@juanmitaboada.com> Juanmi Taboada <juanmi@centrologic.com>
```

br0th3r:
Thank you, AUTHORS corrected. :-)

src/escpos/printer.py
```diff
@@ -52,47 +54,38 @@ def __init__(self, idVendor, idProduct, timeout=0, in_ep=0x82, out_ep=0x01, *arg
...
-        except NotImplementedError:
-            pass
+        self.device = dev.open()
+        if platform.system() == 'Linux' and self.device.kernelDriverActive(self.INTERFACE):
+            self.device.detachKernelDriver(self.INTERFACE)
 
-        if check_driver is None or check_driver:
-            try:
-                self.device.detach_kernel_driver(0)
-            except usb.core.USBError as e:
```
patkan: Would it be possible to also check for this exception in your code?

setup.py
```diff
@@ -107,7 +107,7 @@ def run_tests(self):
         'Topic :: Office/Business :: Financial :: Point-Of-Sale',
     ],
     install_requires=[
-        'pyusb>=1.0.0',
+        'libusb1',
```
patkan: Are all versions of this lib OK?

br0th3r: I haven't checked other versions of libusb1.

src/escpos/printer.py
br0th3r: I am not sure if this is necesary with libusb1.

br0th3r:
Pyusb didn't work on Windows for me after 10h trying on several ways. Reading on Internet I got to know that there several differences between the API v0.1 and v1.x of libusb1 and Pyusb is not compatible with the latest which happens to work properly on Windows.

I have checked this library on Linux and Windows 7 for an EPSON TM-T20II (MODEL M267D) and it worked like a charm. :-)

patkan:
So if nobody opposes I would merge this in the coming days.

belono:
Is possible to extend support for libusb1 without dropping pyusb?

patkan:
I will research how to set optional dependencies for a Python package. This way we could use both libraries.

What bothers me a bit now is that I think I remember that pyusb allegedly supports libusbv1.
I will also research this.

patkan:
It seems possible to specify platform-dependant requirements. https://stackoverflow.com/a/32955538/4244236 https://stackoverflow.com/a/42946495/4244236
This change should keep the old library an allow to switch to the new dependency. Ideally the proper package is preselected for the platform and can be switched. (In the module this can be realized by trying to import the first one, then falling back to the other if the import does not succeed.)

om26er:
@patkan @belono I made an attempt to make things work on Windows here https://github.com/python-escpos/python-escpos/pull/330

PR #289, opened 2018-04-09
parameter for implementation of nonnative qrcode

primax79:
### Contributor checklist

 * Custom Q3X

----------

### Description
Custom printers have problems printing images in bitImageRaster, sometime they print 2 times the image, or a piece of it.
They don't support native qrcode, so I needed to print nonNative qrcode with bitImageColumn implementation

I just exposed the impl paramin qr, to be used in nonnative implementations


patkan:
Thank you for the PR!

PR #311, opened 2018-09-21
Allow arbitrary USB args

gmarull:
### Contributor checklist

 * POS-5890

----------

### Description

This change allows to pass arbitrary arguments to `usb.core.find`. This can be useful, for example, to pass a custom matcher and therefore be able to instantiate 2 printers with the same vid/pid connected on the same PC.

Note that public API is left unchanged.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -8,6 +8,7 @@ Curtis // mashedkeyboard
 Davis Goglin
 Dean Rispin
 Dmytro Katyukha
+Gerard Marull-Paretas
 Hark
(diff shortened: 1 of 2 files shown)


patkan:
Thank you for the PR! Looks good to me. I will try it out asap, but do not see any obstacle for merging.

patkan:
Thank you very much for your contribution and sorry for taking so long :-)

gmarull:
Thanks!

Gerard Marull-Paretas
teslabs.com

Issue #176, opened 2016-09-06
ideas for api changes


patkan:
Hi,
we will soon be releasing version 3. Which aspects of the API should be changed until then and how?

miracle2k:
The capabilities patch already contained an API change - the constructor no longer takes the number of profiles.

Personally, I would take more time to collect a bunch of API changes and then release.

patkan:
Jep, me too. I just wanted to collect what has to be done.
I think we definitely have to improve set().

Apart from that I would like to improve the style of parameters. Currently we have method-parameters like `type="BOLD"`. I would prefer something like `type=texttype.bold `. In order to make it somewhat compatible we could set `texttype.bold="BOLD"`. This way we can do string comparison and even catch the old parameters.

And maybe we can find a clean structure that when you do `import escpos` you get everything what you need for using python-escpos.

mike42:
I agree with the issues with `set()` noted above.

Would you consider an implementation which separates `set()` into small single-purpose methods, uses something other than string literals as arguments, and provides a replacement `set()` to maintain some continuity?

Eg, maybe the syntax is off, but to illustrate in code:

```python
class FontUnderline:
      None, Single, Double = range(3)

class FontWeight:
      None, Bold = range(2)
```

```python
class Printer:
...
    def setUnderline(underline=FontUnderline.None):
...
    def setWeight(weight=FontWeight.Normal):
...
    def set(**kwargs):
        # Just read arguments and delegate
...
```

And the user can still:

```python
foo.set(weight=FontWeight.Bold, underline=FontUnderline.Double)
```

Or..

```python
foo.setWeight(FontWeight.Normal)
```

This could help keep complexity down so that `set` doesn't get out of hand as more features are added.

miracle2k:
My thoughts: 

- I'm not a fan of a ``set(type=TextType.Bold)``-style API. That's a Java-ism to me. There isn't any benefit in Python compared to a string: ``set(type="bold")``. Only now I have to import extra types (and google the naming of TextType).

- I do like staying close to raw ESC/POS. The `text_type` argument we currently have is our own invention, I don't see the need for that. Just let the user set "bold" and "underline" separately.

- I like simplicity. Boldness is a boolean, so I would suggest `set_bold(True)`. If the term used were "weight", I would still prefer ``set_weight("bold")``. For underline, which has single and double options, I would support: ``set_underline(True|False|single|double)``. 

[...]

PR #339, opened 2019-06-15
release v3.0a5

patkan:


om26er:
Thank you for the release and subsequent pre-release update on pypi.

om26er:
Also, I have just found https://github.com/python-escpos/python-escpos/milestone/5 -- is that *all* that needs to be fixed before 3.0 stable release ?

Either way, I will be helping to get 3.0 released as stable.

patkan:
Thank you for asking :-)
I will go through the issues and try to see if there should be anything added from my point of view. Basically I would like to include all current changes of the API in v3. On the other hand, release numbers are not made of gold and should we need a v4, we can have a v4 ;-)

In #176 we tried to discuss what could be changed. The other issues are more something that just has to be done.

patkan:
As far as I see it, the milestone should be cleaned up now, @om26er. In case you would like to work on any issue, feel free to pick it up :-)

PR #356, opened 2019-08-03
Add Chinese support

NullYing:
### Contributor checklist

----------

### Description

A rougher way to support Chinese, maybe there is a better way?


patkan:
Hi, thank you for your contribution.
May I ask you please to rebase your commit messages to English?

Is this method really necessary? As far as I see it, the magic encode feature should handle this transparently. If not, the best place to fix this would be in the magic encode module.

NullYing:
> Hi, thank you for your contribution.
> May I ask you please to rebase your commit messages to English?
> 

Sadly,  magic coding module cannot handle Chinese

patkan:
> Sadly, magic coding module cannot handle Chinese

Yes, and that is why it should be patched there :-)
I think it is not necessary to change the interface of the text function in order to achieve this.

NullYing:
Now, the magic encode feature will handle this transparently

patkan:
Hi, thank you for the changes. I took the liberty to update the mailmap file. The authors-check is a bit broken when it goes to unicode-handling and with the mailmap file this seems to be fixable.

I will have to think about this whether this collides somehow with any "normal" codepages. Also do you have to issue a certain command in order to set the printer to accepting chinese characters? If this does not collide with anything, I would say we can merge this and later on improve it if necessary (I think there is no "standard" way of printing chinese characters, so maybe different printers have to be handled differently)

@mike42 You probably know better the encoding handling of those printers. What do you think, is there any issue merging these changes?

mike42:
I have not actually run the code, but on inspection I think that it's fine to merge, simply because text which doesn't contain CJK characters will be largely unaffected. As far as I know, we have no existing support for anything in that Unicode block, it would just print a lot of "?".

I will point out that this does not segment the text into CJK and non-CJK parts. After this change, non-Chinese, non-ASCII characters will be sent out as GB 18030 based on whether a Chinese character was used elsewhere in the text, which I am guessing is not correct. Eg. compare the binary output from these test strings to observe how differently they are encoded:

- "木 mù tree…" (uses GB 18030 for the whole string)
- "mù tree…" (uses CP437, followed by a different, printer-dependent encoding for "…")

We would ideally also check for support in the printer database before using the encoding, since it's one of many schemes for adding Chinese text to printers, and you will get gibberish if you use an unsupported encoding.

I would generally recommend against merging code which may print gibberish, because arbitrary binary can trigger commands, and that could cause all kinds of trouble. In this case however, [Wikipedia](https://en.wikipedia.org/wiki/GB_18030#Mapping) suggests that this particular encoding will not contain non-printables, so it looks quite safe.

Oyami-Srk:
If a string contains text that can be decoded by gb18030 and text can't, Will this simple manipulation break the other text which cannot be decoded by gb18030?

NullYing:
> If a string contains text that can be decoded by gb18030 and text can't, Will this simple manipulation break the other text which cannot be decoded by gb18030?

yes, but this situation is rare

DroneMesh:
Use Zadig.exe to change the driver of the printer to libusbk and then you can detect it and print with it.

PR #366, opened 2019-10-07
Drop EOL Python 2.7 support, use python-barcode instead of viivakoodi

MicroJoe:
### Contributor checklist

 * POS-5890

----------

### Description

Python 2.7 [is reaching its end of life on 2020-01-01](https://pythonclock.org/). We should embrace the extinction of the 2.x Python releases, which will lead to unified codebase, and less work for all of the Python maintainers in the world.

This also allows us to introduce [python-barcode](https://github.com/WhyNotHugo/python-barcode) as a replacement for [viivakoodi](https://github.com/kxepal/viivakoodi). It is a maintained fork of pyBarcode that is focusing on Python 3 only, so that double version support code can be dropped, and code can be cleaned. At least [one new codebar type is implemented](https://github.com/WhyNotHugo/python-barcode#v0100) (GS1-128). However, minimal version supported by [this best choice for Python barcode generation](https://jonemo.github.io/neubertify/2019/01/23/barcode-generation-python/) is Python 3.5, so this PR is effectively removing Python 2.7 and 3.5 as a side effect.

Accepting this PR before python-escpos before final 3.0 release allows us to not care about Python 2.7 for this release. On the funny side, it would be nice for python-escpos 3.0 release to only support 3.x Python release, too.


setup.cfg
```diff
@@ -2,10 +2,6 @@
 verbosity=3
 with-doctest=1
 
-[bdist_wheel]
-# This flag says that the code is written to work on both Python 2 and Python 3.
-universal=1
```
patkan, line 7: Are you sure that this flag has no relevance for a Python3-wheel?

MicroJoe: I am not sure about this, I did this change based on the comment itself. It should be investigated, but IIRC I tried this change with Python 3 and it worked just fine. What is a test scenario that we could try to see if this line is still relevant?

MicroJoe:
> Hi @MicroJoe, good to hear from you again! (And as always sorry for taking a while :-P )
> Basically, we can do this now. I have just a small question.
> 
> All in all, I'll try to clean up the code and finish the transition to a release 3.

Hey @patkan nice to hear from you too! This change is a bit strange because it both remove support for Python 2 *and* change the barcode implementation library. However the two commits can be separated if required so that we first focus on Python 2 removal and then start thinking about using this new barcode library.

What do you think?

Also, I see that you merged the `development` branch into my branch, maybe I should just rebase it upon development so that it will make the merge easier. There will be some conflicts maybe, but I can have a look. Or do you prefer to keep the merge you did and I do not touch my branch?

MicroJoe:
Let me abandon this PR and create a new one for Python 3 only migration.

patkan:
Hi @MicroJoe we could habe also done this in one step, but it's also OK if we separate it.

I have merged in the development branch as we used the master just for tracking the releases. But  normally that is what we use thereleases tags for. This added unnecessary complexity. (Which I am trying to remove)

PR #392, opened 2020-05-10
drop Python 2.7 support

### LINKED ISSUE #371: Drop python2 support
This needs to happen sooner or later since Python 2 is more or less getting EOL on 31st December (There will be a final Python 2.7 release in April).

If this is agreed, I can do the change.

MicroJoe:
### Contributor checklist

 * POS-5890

----------

### Description

Python 2.7 [is reaching its end of life on 2020-01-01](https://pythonclock.org/). We should embrace the extinction of the 2.x Python releases, which will lead to unified codebase, and less work for all of the Python maintainers in the world.

Accepting this PR before python-escpos before final 3.0 release allows us to not care about Python 2.7 for this release. On the funny side, it would be nice for python-escpos 3.0 release to only support 3.x Python release, too.

Closes #371.


MicroJoe:
Great, I see that you have found the codecov problem as I did. I was not able to fix it but was able to correct the flake8 issues. Thanks in advance for the merge, after this I will propose a new PR for the barcode upgrade.

PR #395, opened 2020-05-10
replace viivakoodi with python-barcode, center soft barcodes by default

MicroJoe:
### Contributor checklist

 * POS-5890

----------

### Description

This introduces python-barcode as a replacement for viivakoodi. It is a maintained fork of pyBarcode that is focusing on Python 3 only, so that double version support code can be dropped, and code can be cleaned. At least one new codebar type is implemented (GS1-128).

It also introduces the possibility to center the barcode, something that is already possible with `qr_code` function. It will just pass the `center` option to the `image` function, but I decided to make it `True` by default because I do not see any reason why the barcode should *not* be centered on the paper.

Also, the barcode-related tests has been updated a little bit.


patkan:
Looks good to me. Thank you!

PR #485, opened 2022-07-05
Add support for slip/cheque dot matrix printers

kedare:
### Contributor checklist

- [x] I have tested my contribution on these devices: NCR 7167 (waiting to have one available to test)

----------

### Description
This PR add supports for printers that have both thermal printing module and a second dot matrix printer (usually for cheque, billing documents, carbon copy, etc...)

The same usual ESC/POS instructions are used to print to both printers, only some specific commands are also available for the slip printer as it works differently.

I am just not sure of what is the correct naming for the new function, if someone can take a look (I was not sure it if was better to prefix them with `slip_` for example or just more proper english `action_object_etc...` like how I done it ?

I did not test the feature yet, I should have a printer available for testing end of the week. (NCR 7167), the ECSPOS commands I used are the ones from the NCR documentation, it looks like it's standard ESCPOS so it should work fine on others models like EPSON, etc...)

https://kedare-public.s3-website.fr-par.scw.cloud/docs/NCR-ESCPOS.pdf

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,9 +1,11 @@
 Ahmed Tahri
 akeonly
+Alejandro Hernández
 Alexander Bougakov
 Alex Debiasio
 Asuki Kono
 belono
+B. Howell
 Brian
 Christoph Heuel
 Cody (Quantified Code Bot)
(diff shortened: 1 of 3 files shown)


kedare:
So the implementation is working fine so far on my NCR printer :)
I just could not get graphics to be printer (except native barcode) but I am not even sure this is supported to work (I can only print via bitColumn on the thermal mode for example).
Let me know if the current implementation and naming looks ok and then I will start the writing the tests (that should be quite easy)

PR #527, opened 2023-05-31
New high level barcode method. Closes #245, #244.

### LINKED ISSUE #244: Get rid of font selection for barcode method
I have:

### Bug description
The font selection for the `barcode` method seems cumbersome to me. Based on the type of the barcode the user wants to use we should be able to automatically detect the correct font to use for the barcode.

I can try to implement this if one agrees that the intelligence should be in the library and not in the user trying to get code128 to work (or maybe I missed something that makes font useful to pass to the method).

### LINKED ISSUE #245: Merge software and hardware barcodes to a same method
I have:

### Bug description
This issue continues discussion started in #221 about barcode generation.

Currently the hardware and software barcodes are implemented in two separated methods for the sake of simplicity. However we want to merge the user call to the `barcode` method, with an optional boolean argument named `force_software=False` than allow users that have hardware barcode capabilities to force software rendering.

If your printer does not support hardware rendering according to the printer database then software barcodes are going to be used automatically.

belono:
### Contributor checklist

- [X] I have tested my contribution on these devices:
 * Epson TM-H6000

----------
Closes #245
Closes #244
Two steps closer to the v.3.0 milestone!!

### Description

This patch adds functionality to allow the `barcode()` method to choose the appropriate bar-code printing implementation (hardware or software), depending on the capabilities of the printer and the type of bar-code to be printed (EAN13, CODE128 , etc.).

The following changes/additions were made:

1- New high level `barcode()` method:

- **Backwards compatible** with the old `barcode()` method.
- **Translates received params** to those needed by the low level methods.
  Addresses, for example, the different bar-code naming conventions (GS1-128, gs1_128), or the different units of measure of the module size among the methods _hw_barcode (points) and _sw_barcode (mm).
- **Attempts to match the size of a native hardware bar-code** and a software-rendered bar-code image.
  For this to be successful, the library needs to know the DPI of the printer to match the bar-code image resolution.
  In the near future, the profile database should include this value in a media dpi field but, until that, the value is calculated, if unknown, from the number of points per line and the printable area width of the printer. See `_dpi()`.
- **Switches to software bar-code generation if hardware is not supported**, either because the printer does not support bar-code printing natively or does not implement the specified bar-code type.
 - **Software bar-code generation can be still forced** by including the param `force_software`, either specifying the image implementation of the printer or letting the method to choose one, according to the capabilities profile.

2- Previous `barcode()` method renamed to `_hw_barcode()`:

- Non public method.
- **Automatic selection of the function_type** value (A or B) if not specified, according to the capabilities profile and the bar-code type.

3- Previous `soft_barcode()` method renamed to `_sw_barcode()`:

- Non public method.
- **The `_sw_barcode()` method now sets additional parameters** and properties to `barcode.ImageWriter()` to try to match the software bar-codes to the size of native hardware bar-codes.

4- New `_dpi()` method:

- Non public method.
- **Looks up a media dpi field** in the capabilities database.
  If not found, **calculate the printer’s dpi** based on the media width data of the printer profile.
  As this calculation is inaccurate due to the different horizontal margins between models, the recommendation is to force the media dpi value manually in the profile.
  In execution time this can be done by extending the profiles dictionary before calling `barcode()`: `printer.profile.profile_data["media"]["dpi"] = 180  # Or the DPI value of your printer`
  Note that this is only in case the printer profile does not contain the media dpi field previously.

### Output
This is the preliminary output of some supported barcode types on an Epson TM-H6000 180 DPI printer.
Please, ignore the small size of the payload text under the sw_barcodes as it is already fixed.
Note the matching sizes of the ean13 samples, and the little larger width of the code39 and code128 ones. This is due to the addition of the check digit 'W' and the charset selector '{B' symbols by the `python-barcode` library that ends up rendering different (larger) barcodes than the native ones.
I don't like this behavior but, since it is not related to this library but mainly to the `python-barcode` library, I think it is out of the scope of this patch to try to fix it, at least at this early stage.

s_ean13_hw_sw
s_code39_hw_sw
s_code128_hw_sw

So, to sum up:
- The patch works well on a 180 dpi TM-H6000 printer, let's see other models.
- It is highly recommended that the printer matches a full featured capabilities profile.
- To print hardware and software barcodes of the same size, we need at least a profile containing the media width in points and mm.
- In a future we want to add a new dpi field into the capabilities database containing each printer's value.  We are already supporting that dpi field in advance.

Waiting for your feedback, questions, suggestions, commits...

Greetings!!


belono:
I've been thinking about the _hw_barcode() and _sw_barcode() being 'private' methods.
My idea was that the user would not have to choose between three *barcode() methods, but it may not be consistent with the rest of the API to 'hide' these methods. Even so, I would like to make it clear that the recommendation is `barcode()`.

Should I remove that leading underscores from those two method names and add some kind of clarification to the *_barcode() docstrings? Or leave it as is?

src/escpos/escpos.py
```diff
@@ -411,6 +424,24 @@ def check_barcode(bc, code):
...
+    def _dpi(self) -> int:
+        """Printer's DPI resolution."""
+        try:
+            dpi = int(self.profile.profile_data["media"]["dpi"])
+        except (KeyError, TypeError):
+            # Calculate the printer's DPI from the width info of the profile.
+            try:
+                px = self.profile.profile_data["media"]["width"]["pixels"]
+                mm = self.profile.profile_data["media"]["width"]["mm"]
+                mm -= 10  # paper width minus margin =~ printable area
```
patkan, line 436: Would we expect this margin to be always (for every printer) 10?

src/escpos/escpos.py
```diff
@@ -422,6 +453,128 @@ def barcode(
...
+
+        :param pos: text position (ABOVE, BELOW, BOTH, OFF) relative to the barcode
+            (ignored in software renderer).
+            *default*: BELOW
+
+        :param font: select font A or B (ignored in software renderer).
+            *default*: A
+
+        :param align_ct: If *True*, center the barcode.
+            *default*: A
```
patkan, line 492: Default true?

patkan (commented):
Thank you very much for the changes and the thorough documentation 😊
I didn't find the time yet to try this out, but it looks good to me.
Maybe you could already create issues in python-barcode (to discuss those added symbols) and escpos-printer-db (to discuss the addition of DPI).
I have added a few minor comments and questions and will add any further remarks as soon as I manage to try your changes.

I would keep the SW and HW methods private like you suggested.

src/escpos/escpos.py
belono: Oops! Yes.

src/escpos/escpos.py
```diff
@@ -422,6 +453,128 @@ def barcode(
...
+
+        :param force_software: If *True*, force the use of software barcode renderer from profile.
+            If *"graphics", "bitImageColumn" or "bitImageRaster"*, force the use of specific renderer.
+        :type force_software: bool | str
+
+        :raises: :py:exc:`~escpos.exceptions.BarcodeCodeError`,
+                 :py:exc:`~escpos.exceptions.BarcodeTypeError`
+
+        .. tip::
+            Get all supported formats at:
```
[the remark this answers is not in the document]
belono: Maybe replace the broken references with a table containing the full barcode types that are supported, ignoring the render method, is better. And additionally change the label ``tip`` to ``note``?

```
| Supported barcode formats |
| :---: |
| UPCA |
| UPCE |
| EAN13 |
| EAN8 |
| CODE39 |
| ITF |
| NW7 |
| CODABAR |
| CODE93 |
| CODE128 |
| GS1_128 |
| GS1_DATABAR_OMNIDIRECTIONAL | GS1_DATABAR_TRUNCATED |
| GS1_DATABAR_LIMITED |
| GS1_DATABAR_EXPANDED |
| EAN |
| EAN13GUARD |
| EAN14 |
| EAN8GUARD |
| GS1 |
| GTIN |
| ISBN |
| ISBN10 |
| ISBN13 |
| ISSN |
| JAN |
| PZN |
| UPC |
```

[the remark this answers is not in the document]
patkan: Sounds good :-)

src/escpos/escpos.py
patkan: Thank you, then we can close this here.

src/escpos/escpos.py
```diff
@@ -422,6 +453,128 @@ def barcode(
...
+        if not any([*capable_bc.values()]):
+            raise BarcodeTypeError(f"Not supported or wrong barcode name {bc}.")
+
+        if force_software or not capable["hw"] or not capable_bc["hw"]:
+            # Select the best possible capable render mode
+            impl = capable["sw"][0]
+            if force_software in capable["sw"]:
+                # Force to a specific mode
+                impl = force_software
+            print(f"Using {impl} software barcode renderer")
```
patkan, line 544: I would prefer for messages like this a logging facility, but as of now we do not have it in python-escpos and therefore I would just keep it as it is and rework all print statements at a later point.

src/escpos/escpos.py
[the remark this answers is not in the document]
patkan: I have thought a bit more about it: how can we keep this list up to date? In the end this changes when python-barcode gets more support.

Please have a look at https://python-escpos--527.org.readthedocs.build/en/527/api/escpos.html#escpos.escpos.Escpos.barcode. If you think this is OK, we could merge it like this, otherwise feel free to add the list as proposed.

patkan:
I have played around with the software rendering and hardware rendering and I really like the implementation. Thank you!

Please have a look if my change is OK for you, then we could close the discussion and merge this. Otherwise feel free to change it.

If the discussions are resolved, I would merge it.

src/escpos/escpos.py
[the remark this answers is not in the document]
belono: I think that's great. In fact, that's what I was trying to get from the beginning, but I didn't know how to do it properly.

belono:
> I have played around with the software rendering and hardware rendering and I really like the implementation. Thank you!

My pleasure!

Everything is OK on my side.

patkan:
Ok, great, than I will merge now

PR #534, opened 2023-07-22
Remove sleep when quering status. Fixes #407

### LINKED ISSUE #407: status reads very slow on Serial printers
Hi, I came across an issue where I found my application stalling when reading printer statuses. Looking at the sources I found that there is a 1 second sleep between the write and the read for status requests. I guess it was meant to give time to the printer to answer back. I understand the motivation, but in practice the same effect of "giving time" is achieved by the timeout parameter in when constructing the Serial() object, with the benefit that the program would only pay the price of the delay (up to 1 second delay in this case) if the printer actually took that time to respond. In practice this is rarely the case, and the reads return almost immediately, with the expected data.

Removing the sleep did the trick for me (reduced printing times by many seconds each time), maybe you could review the current approach and make this change permanent.

Thanks.
Regards.
Eduardo

belono:
### Description
Remove the unneeded `sleep()` that stalls the printer status query.
Fixes #407.

### Tested with

### DIFF
--- a/src/escpos/escpos.py
+++ b/src/escpos/escpos.py
@@ -1174,7 +1174,6 @@ class Escpos(object):
         :rtype: array(integer)
         """
         self._raw(mode)
-        time.sleep(0.1)
         status = self._read()
         return status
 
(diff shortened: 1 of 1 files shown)


patkan:
Looks good to me

PR #548, opened 2023-08-21
fix typo in lp connector

belono:
### Description

Typo at the function `is_usable()` of the module `lp` :
platform has `not` to be windows

### Tested with


patkan:
Great find, thank you :-)

PR #554, opened 2023-08-28
Separate method open() and constructor  Fixes #445

### LINKED ISSUE #445: Separate method open() and constructor
### Bug description
Any implementation of retrying algorithm makes code complicated

### Steps to reproduce
i.e. `tenacity` package

In device implementation method open() makes it impossible to implement retrying policy if device is not connected since instance is created after method is called. Don't you think that logic of creating object and connecting to device must be separated? 

In this case there will be no recursion errors and easy implementation than self-made logic 
```
def discover(self):
        while not self.device:
            try:
                self.device = Usb(idVendor=int(os.environ.get('PRINTER_VENDOR_ID'),16),
                                    idProduct=int(os.environ.get('PRINTER_PRODUCT_ID'),16),
                                    in_ep=int(os.environ.get('PRINTER_IN_EP'),16),
                                    out_ep=int(os.environ.get('PRINTER_OUT_EP'),16),
                                    timeout=int(os.environ.get('PRINTER_TIMEOUT')))
                return self.device
            except Exception as e:
                logger.error({'message':repr(e)})
                time.sleep(0.5)
                continue
```
Expected 
```
class MyUsbPrinter(Usb):
    def __init__(self):
       super().__init__(idVendor=int(os.environ.get('PRINTER_VENDOR_ID'),16),
                                    idProduct=int(os.environ.get('PRINTER_PRODUCT_ID'),16),
                                    in_ep=int(os.environ.get('PRINTER_IN_EP'),16),
                                    out_ep=int(os.environ.get('PRINTER_OUT_EP'),16),
                                    timeout=int(os.environ.get('PRINTER_TIMEOUT')))
    def connect(self):
        r = retry(**kwargs) 
        usb_args = {}
        usb_args['idVendor'] = self.idVendor
        usb_args['idProduct'] = self.idProduct
        r(self.open, usb_args={})(self)

my_printer = MyUsbPrinter()
my_printer.connect()
````
And also can be used as callback for read() and write() operations if connection is down to prevent application break

```
def write(self):
    r = retry(**kwargs, retry_error_callback=self.reconnect(self), retry=retry_if_exception_type(USBError))
    # write to device
   r(self._raw, data, timeout)
````

Discuss if it fits  logic, please. Another way is to implement Usb device without inheritance.

### PR DESCRIPTION
**Pull request by belono**
### Description
Separate the method `open()` from the class constructors and enhance consistency between connectors.

This patch fixes #445 **removing** the call to the `open()` method at the end of the printer's `__init__()` constructor and **converting** to properties the necessary variables. This facilitates the reuse and reconfiguration of the object to simplify the implementation of retry policies or other usages, as @rsaleev and @drakorg requested, but trying to keep some compatibility with the prior implementation thanks to the addition of an optional open-on-first-call mechanism, so the users don't need to refactor their old code if they don't need this features.

This PR will consist in:
- [x] Decouple object creation and connection, keeping some backwards compatibility.
- [x] Enhance consistency between connectors.

At separated PRs:
- [ ] Document.
- [ ] Write examples.
- [ ] Write tests.

### Tested with
ESC/POS Printer Simulator

### CODE REVIEW

File: src/escpos/escpos.py [hunk 1]
```diff
@@ -128,6 +128,27 @@ def __del__(self):
         """Call self.close upon deletion."""
         self.close()
 
+    @property
+    def device(self):
```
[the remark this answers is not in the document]
belono: I'm currently adding type annotations to the printer constructors for the next batch of commits ("Enhance consistency between connectors") where all printers will have default init values, but honestly I'm having some doubts about the correct type of some params like `in_ep` and `out_ep` of the `Usb()` class or `parity` and `stopbits` of the `Serial()` class.
We might review those ones and others when I push the changes.

src/escpos/printer/cups.py
```diff
@@ -20,6 +20,9 @@
     import cups
 
     _DEP_PYCUPS = True
+    # Store server defaults before further configuration
+    DEFAULT_HOST = cups.getServer()
```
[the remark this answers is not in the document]
belono: Yes, it is safe.
`cups.getServer()` and `cups.getPort()`have internal defaults so a value is always returned if the `pycups` library is successfully imported.

In fact, I've successfully disabled the cups service on my machine and then import and instance a `CupsPrinter()` object .
Of course it will raise an exception after you send a print command but that's the expected behavior.

src/escpos/escpos.py
```diff
@@ -128,6 +128,27 @@ def __del__(self):
...
+        """Implements a self-open mechanism.
+
+        An attempt to get the property before open the connection
+        will cause the connection to open.
+        """
+        if self._device is False:
```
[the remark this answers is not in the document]
belono: Well, `bool(None)` returns `False` but `None` and `False` are different objects. Then `bool(None) == False`, `bool(None) is False` but `None != False` and `None is not False` because python checks if both are the same object, so yes, this way one can differentiate between the two states (objects).

Changing the state of the `device` property from 'not initialized' (False) to 'initialized but not opened' (None) was the way I found to auto-open the connection only on the first call, otherwise, as you guess, every check if the device exists would try to reopen the connection.
A side effect is that when the user instantiates a printer for the first time and then checks the value of `device` then the connection tries to be opened:
```
import escpos.printer as printer

>>> p = printer.Serial()
>>> print(p.device)
Serial printer Enabled
Serial<id=0x7f8342d78790, open=True>(port='/dev/ttyUSB0', baudrate=9600, bytesize=8, parity='N', stopbits=1, timeout=1, xonxoff=False, rtscts=False, dsrdtr=True)
```
but this behavior could be useful to check the connection before sending any commands.

src/escpos/printer/cups.py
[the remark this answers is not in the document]
patkan: Ok, great. Thank you for the details!

src/escpos/escpos.py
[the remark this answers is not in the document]
patkan: I agree. Is there then a way to restore the connection? (Apart from explicitly calling open()?)

src/escpos/printer/file.py
```diff
@@ -73,6 +72,8 @@ def _raw(self, msg):
...
     def close(self):
         """Close system file."""
-        if self.device is not None:
-            self.device.flush()
-            self.device.close()
+        if not self._device:
+            return
+        self.device.flush()
+        self.device.close()
+        self.device = None
```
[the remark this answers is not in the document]
belono: I agree it makes more sense to restore the connection to False (Not initialized) after `close()`.
As far as I remember, the only reason to set `device` to None on `close()` was because we were checking `if self.device is None` in some places but I changed those checks to `if not self.device` so either False or None will pass the condition.
I missed that. Thanks!

src/escpos/escpos.py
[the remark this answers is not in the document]
belono: After commit 7103af083c2628006d6d1b53908f6dcfcc2b433d a closed connection will be reopened in the same way as the first time (self-open or explicit call to `open()`), which allows the reuse and|or reconfiguration of the object.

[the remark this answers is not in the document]
patkan: Great, I think this is resolved then

src/escpos/escpos.py
```diff
@@ -14,7 +14,7 @@
 import textwrap
 from abc import ABCMeta, abstractmethod  # abstract base class support
 from re import match as re_match
-from typing import List, Optional, Union
+from typing import List, Optional, Union, Literal
```
patkan, line 17:
```suggestion
from typing import List, Literal, Optional, Union
```

src/escpos/printer/serial.py
```diff
@@ -10,6 +10,7 @@
 
 
 import functools
+from typing import Union, Optional
```
patkan, line 13:
```suggestion
from typing import Optional, Union
```

patkan (commented):
This should please the sorting tool (isort)

src/escpos/escpos.py
```diff
@@ -128,6 +132,27 @@ def __del__(self):
         """Call self.close upon deletion."""
         self.close()
 
+    @property
+    def device(self) -> Union[Literal[False], Literal[None], object]:
```
patkan, line 136:
```suggestion
    def device(self) -> Union[Literal[None], object]:
```
I think the return value cannot become False.

src/escpos/printer/serial.py
belono: We've fixed it at the same time. ;-D

src/escpos/escpos.py
belono: mypy cli doesn't complain about `def device(self) -> Union[Literal[False], Literal[None], object]:`

I think `False` is not accepted but it is `Literal[False]`

patkan: I just meant with my comment, that we could narrow down the types that are reported here. If I have a look at the code, I cannot see any way that there would ever be a literal False returned. Therefore we could "add" that as information.

belono: I misunderstood your comment, sorry.
You are right, device only returns None or the connection object.
Thank you very much!

patkan: Thank you! 👍

belono:
Here we go with another little batch of commits. This time there is enhanced consistency between printer connectors what means that I tried to make all the connectors (`escpos.printers`) have **similar** behavior, despite each of them use a completely different library (pyusb, pyserial, socket, win32print, etc...) to communicate with the printer. Also, there is better code consistency and stability thanks to the addition of type checking and more exception catching.

With this new implementation:

-  All the connectors have their `__init__()` parameters defaulting to empty values so the classes can be optionally instantiated without parameters.

-  The `open()` methods gained a new `raise_not_found` parameter defaulting to `True` that raises a new `DeviceNotFoundError` exception if the connection failed to establish.
 In the other hand, `raise_not_found=False` logs an error message and gracefully cancels the opening avoiding to stop the execution.
-   `open()` will `close()` the connection if not closed before, flushing commands or print jobs to the printer on some connectors.
-  `Win32Raw()` gains a new `printers` property returning the available printers. Similar functionality to that one in `CupsPrinter()`.
-  `Usb()`’s `open()` method has been split to improve exception catching.
-  `printer` modules console outputs have been slightly improved and initially moved to `logging` (advancing #559 ).
-  `mypy` type checking compliant.

In addition to the basic usage, these changes allow for some advanced usages like, i.e:
Sub-classing a context manager, re-configuring, and testing a connection using the [tenacity retrying library](https://github.com/jd/tenacity) package.

```
import escpos
import escpos.printer as printer
from tenacity import retry, stop_after_attempt, wait_fixed


class MyPrinter(escpos.escpos.EscposIO):
    """Subclass of EscposIO to allow for context managing."""

    def __init__(self, connector):
        super().__init__(connector)

    @retry(wait=wait_fixed(2), stop=stop_after_attempt(3))
    def connect(self):
        """Test connection.

        Raise tenacity.RetryError on fail.
        """
        print("Connecting...")
        try:
            self.printer.open()
        except escpos.exceptions.DeviceNotFoundError:
            print("Connection error")
        self.printer.is_online()


kwargs = {"devfile": "fake/device"}
connector = printer.Serial(**kwargs)
with MyPrinter(connector) as my_printer:
    # Reconfiguration of the connector object through the ctx_mgr printer property.
    my_printer.printer.devfile = "/dev/ttyUSB0"
    my_printer.connect()
```
Note that the EscposIO context manager has some rough edges with the new behavior of the `device` property that have to be polished, so take this example as a proof of concept.

src/escpos/printer/cups.py
```diff
@@ -78,7 +84,7 @@ def is_usable() -> bool:
         return is_usable()
 
     @dependency_pycups
-    def __init__(self, printer_name=None, *args, **kwargs):
+    def __init__(self, printer_name: str = "", *args, **kwargs):
```
patkan, line 87: You can remove `:type printer_name: ...` from the docstring if you have set it with a type annotation

belono: Done.
Resolving.

belono:
The fix of an ugly bug in the `LP` printer has ended with that connector receiving some love.

Commit f8deb69f43d2a4e2d67873c1f71bdabfaab22ab1 includes:
- Fix error catching implementation.
- Add new property `printers` which returns the system's available printers.
- Add the optional `job_name` parameter to the `open()` method to name the print job.
- Add method `_get_system_default_printer()` that returns the system default printer name.
 
So, the `LP` printer now has the same features as the other high-level connectors (`Win32Raw` and `CupsPrinter`).

belono:
More love to the `LP` printer and some to the `File` printer too:

- Fix `LP` wasting paper due to `auto-flush` enabled by default.
- `LP` and `File` printers flush commands on close if not `auto-flush`.
- `File` printer now calls its own `flush()` method on close to facilitate testing.

src/escpos/printer/cups.py
patkan, line 1: Return types for send and close are None?

src/escpos/printer/file.py
```diff
@@ -73,6 +96,10 @@ def _raw(self, msg):
 
     def close(self):
```
patkan, line 97:
```suggestion
    def close(self) -> None:
```

src/escpos/printer/file.py
```diff
@@ -48,14 +51,34 @@ def __init__(self, devfile="/dev/usb/lp0", auto_flush=True, *args, **kwargs):
...
+            if raise_not_found:
+                raise DeviceNotFoundError(
+                    f"Could not open the specified file {self.devfile}:\n{e}"
+                )
+            else:
+                logging.error("File printer %s not found", self.devfile)
+                return
+        logging.info("File printer enabled")
 
     def flush(self):
```
patkan, line 83:
```suggestion
    def flush(self) -> None:
```

src/escpos/printer/lp.py
```diff
@@ -61,39 +62,106 @@ def is_usable() -> bool:
...
+                raise DeviceNotFoundError(
+                    f"Unable to start a print job for the printer {self.printer_name}:"
+                    + f"\n{e}"
+                )
+            else:
+                logging.error("LP printing %s not available", self.printer_name)
+                return
+        logging.info("LP printer enabled")
 
     def close(self):
```
patkan, line 148:
```suggestion
    def close(self) -> None:
```

src/escpos/printer/lp.py
```diff
@@ -61,39 +62,106 @@ def is_usable() -> bool:
...
-        self.lp.terminate()
+        if not self._device:
+            return
+        logging.info("Closing LP connection to printer %s", self.printer_name)
+        if not self.auto_flush:
+            self.flush()
+        self.device.terminate()
+        self._device = False
 
     def flush(self):
```
patkan, line 158:
```suggestion
    def flush(self) -> None:
```

src/escpos/printer/network.py
```diff
@@ -90,9 +120,12 @@ def _read(self):
 
     def close(self):
```
patkan, line 121:
```suggestion
    def close(self) -> None:
```

src/escpos/printer/serial.py
```diff
@@ -148,6 +167,10 @@ def _read(self):
 
     def close(self):
```
patkan, line 168:
```suggestion
    def close(self) -> None:
```

src/escpos/printer/usb.py
```diff
@@ -157,6 +192,10 @@ def _read(self):
     @dependency_usb
     def close(self):
```
patkan, line 193:
```suggestion
    def close(self) -> None:
```

src/escpos/printer/win32raw.py
```diff
@@ -70,36 +73,73 @@ def is_usable() -> bool:
...
+                    f"Unable to start a print job for the printer {self.printer_name}:"
+                    + f"\n{e}"
+                )
+            else:
+                logging.error("Win32Raw printing %s not available", self.printer_name)
+                return
+        logging.info("Win32Raw printer enabled")
 
     @dependency_win32print
     def close(self):
```
patkan, line 134:
```suggestion
    def close(self) -> None:
```

patkan (commented):
Looks great to me so far.
Do you need help with extending the tests?

By the way, if you want to participate in [Hacktoberfest](https://hacktoberfest.com/participation/), let me know and we can split the PR in a few logically grouped PRs starting from 1st of October.

Also we should probably make the logging configurable insofar as we should not log into the default logger and allow to pass a logger to the library. But for me that could (and maybe should) be part of another PR.

src/escpos/printer/cups.py
belono: Done!

src/escpos/printer/win32raw.py
```diff
@@ -60,6 +65,8 @@ class Win32Raw(Escpos):
 
     """
 
+    _device: Union[Literal[False], Literal[None], _win32typing.PyPrinterHANDLE] = False
```
patkan, line 68:
```suggestion
    _device: Union[Literal[False], Literal[None], "_win32typing.PyPrinterHANDLE"] = False
```
This should prevent the undefined name

src/escpos/printer/win32raw.py
```diff
@@ -70,36 +77,75 @@ def is_usable() -> bool:
...
-        if printer_name is not None:
-            self.printer_name = printer_name
-        else:
-            self.printer_name = win32print.GetDefaultPrinter()
-        self.hPrinter = None
-        self.open()
+        self.printer_name = printer_name
+        self.job_name = ""
+
+    @property
```
patkan, line 86:
```suggestion
    @property
    @dependency_win32print
```
Since this property is also using win32, we should guard it.

src/escpos/printer/win32raw.py
```diff
@@ -70,36 +77,75 @@ def is_usable() -> bool:
...
+            self.close()
+
+        self.job_name = job_name
+        try:
+            # Name validation, set default if no given name
+            self.printer_name = self.printer_name or win32print.GetDefaultPrinter()
+            assert self.printer_name in self.printers, "Incorrect printer name"
+            # Open device
+            self.device: Optional[
+                _win32typing.PyPrinterHANDLE
```
patkan, line 119:
```suggestion
                "_win32typing.PyPrinterHANDLE"
```
Same here with the import and name.

src/escpos/printer/win32raw.py
belono: I spent yesterday struggling with this issue without success.
Thank you very much for the tip!

src/escpos/printer/win32raw.py
belono: You saved my day ;-D

belono:
Hi @patkan !

I need your help here.

Unfortunately, the `_win32typing` module in the `pywin32` package is inaccessible and cannot be imported, so the attempt to fix type checking in the `Win32Raw` printer connector by including the `"_win32typing.PyPrinterHANDLE"` to the `_device`'s type annotations ended up breaking the printer connector.

Since I couldn't find a way to make it work (maybe due to my ignorance in this matter), I investigated a bit and finally managed to find two different solutions for this situation:

- The simplest and obvious one is to annotate the `_device` property with `typing.Any` and forget about the type.
- The second involves adding a new dependency to the library, the [pywin32-stubs](https://pypi.org/project/pywin32-stubs/) package, which makes pywin32 types and constants accessible, bringing up the type checking without breaking anything.
   I've installed the `pywin32-stubs` package in my local machine and then replaced `import _win32typing` with `import win32helper.win32typing as _win32typing` at the start of the win32raw.py module and it works without any other change.

If you know any other way to fix this issue I'll be glad to hear about it, otherwise let me know which one of the previous solution you prefer, please.

belono:
Issue update:

Well, I'm pretty convinced that the package `types-pywin32` is the cause of all of my last 3 day headaches :rofl: 

This package is causing (apart of the aforementioned inaccessible type annotations) some weird false positives from `mypy`, even after the `_device` property is annotated with `typing.Any`. 

For example, mypy on Windows throws:
`error: Argument 3 to "StartDocPrinter" has incompatible type "tuple[str, None, str]"; expected "int"`
Which is a false positive because `StartDocPrinter` accepts the positional arguments `(hprinter, level, _tuple)` as stated in the [pywin32 docs](https://mhammond.github.io/pywin32/win32print__StartDocPrinter_meth.html) so the 3rd argument has to be a tuple, not an int. Additionally changing the position of the arguments to please `mypy` breaks the module.

Nevertheless, simply replacing the `types-pywin32` package with `pywin32-stubs` fixes all the type checking errors in the `Win32Raw` printer: no more unknown types and no more false positives on Windows or Linux.
So, from my point of view, it is clear that we have to replace them.

belono:
Type checking the win32raw.py module with mypy

When `types-win32` is installed, set `_device` type annotations to `Any`:
```python
from typing import Any, Literal, Optional, Union
.
.
.
def __init__(self, printer_name: str = "", *args, **kwargs):
    .
    .
    .
    self._device: Union[Literal[False], Literal[None], Any = False
```
When `pywin32-stubs` is installed, set `_device` type annotations to `PyPrinterHANDLE`:
```python
from typing import Literal, Optional, Union
import win32helper.win32typing as _win32typing
.
.
.
def __init__(self, printer_name: str = "", *args, **kwargs):
    .
    .
    .
    self._device: Union[Literal[False], Literal[None], "_win32typing.PyPrinterHANDLE" = False
```

Summary:

1. Linux: 
- `types-pywin32` installed (`pywin32-stubs` can't be installed on Linux):
  Test result: `error: Argument 3 to "StartDocPrinter" has incompatible type "tuple[str, None, str]"; expected "int"` - **FALSE POSITIVE**

  _FIX: Uninstalling `types-pywin32` fixes the error._


2. Windows:
- `types-pywin32` installed
  Test result: error: Argument 3 to "StartDocPrinter" has incompatible type "tuple[str, None, str]"; expected "int" - **FALSE POSITIVE**
- Both `types-pywin32` + `pywin32-stubs` installed:
  Test result :  `error: Module has no attribute "PRINTER_ENUM_NAME"` - **FALSE POSITIVE**

  _FIX: Uninstalling `types-pywin32` and installing `pywin32-stubs` fixes the error._

So I'm pushing the necessary changes to disable `types-pywin32` and add `pywin32-stubs` as a required dependency for the win32 platform to help with the review process.
Please don't doubt on share any suggestions or concerns about this change. If I'm ignoring or misunderstanding something I will be happy to make the necessary corrections.

patkan:
Sounds good to me, feel free to switch the type annotations to the other package.
That means we need a mypy run on windows and one on Linux, correct? There are two CI pipelines for both Linux and Windows, so this is no issue but we should do that then.

belono:
Finally I had to remove the `types-pywin32` and `pywin32-stubs` dependencies and its usage in Win32Raw.
The CI pipeline is green again and the module is well enough type-annotated, so I think we could leave it as is for now.

belono:
I moved the platform/dependency checker decorator for platform-specific printers to the `__init__` constructor of its classes.
This should be enough to prevent its use where it is not possible.

patkan (approved):
I approve these changes, thank you very much for your hard work.

If you want to, please close this pull request, and open a new one (with the changes on this branch), mentioning this one for reference. I will merge this new pull request (which will make it count towards Hacktoberfest)

PR #560, opened 2023-09-04
[FIX] typing_extensions is missing

fmdl:
With this commit the module doesn't work : https://github.com/python-escpos/python-escpos/commit/44444f3c5100080c9a635112e592733be402b2c0

### Tested with

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,1 +1,2 @@
 -e .
+typing_extensions


requirements.txt
```diff
@@ -1 +1,2 @@
--e .
\ No newline at end of file
+-e .
+typing_extensions
```
patkan, line 2: This is unfortunately the wrong location for requirements changes. Please place them in setup.cfg

patkan (changes requested):
Thank you for the contribution.
On which python version are you experiencing issues? As far as I know all of the used typing features should be available in >=3.8 (the oldest supported version of this lib)

chrisv2:
I have this problem too when using the `python:3.10-bullseye` docker image. Actually, the culprit is in [`escpos/types/__init__.py`](https://github.com/python-escpos/python-escpos/blob/master/src/escpos/types/__init__.py#L6), where `TypedDict` gets imported explicitly from `typing_extensions`.

Adding `typing_extensions` to `setup.cfg` would solve the issue, however I don't think that is the best approach because

a.) `TypedDict` is already available in Python 3.8 and later, even without typing_extensions
b.) python-escpos does not use any of the backported features from typing_extensions

Suggested fix is to get rid of `typing_extensions` altogether:

```diff
--- a/src/escpos/types/__init__.py
+++ b/src/escpos/types/__init__.py
@@ -1,9 +1,7 @@
 # -*- coding: utf-8 -*-
 """Custom types."""
 
-from typing import Dict
-
-from typing_extensions import TypedDict
+from typing import Dict, TypedDict
 
 
 class ConstTxtStyleClass(TypedDict):
```

patkan:
> I have this problem too when using the `python:3.10-bullseye` docker image. Actually, the culprit is in [`escpos/types/__init__.py`](https://github.com/python-escpos/python-escpos/blob/master/src/escpos/types/__init__.py#L6), where `TypedDict` gets imported explicitly from `typing_extensions`.
> 
> Adding `typing_extensions` to `setup.cfg` would solve the issue, however I don't think that is the best approach because
> 
> a.) `TypedDict` is already available in Python 3.8 and later, even without typing_extensions b.) python-escpos does not use any of the backported features from typing_extensions
> 
> Suggested fix is to get rid of `typing_extensions` altogether:
> 
> ```diff
> --- a/src/escpos/types/__init__.py
> +++ b/src/escpos/types/__init__.py
> @@ -1,9 +1,7 @@
>  # -*- coding: utf-8 -*-
>  """Custom types."""
>  
> -from typing import Dict
> -
> -from typing_extensions import TypedDict
> +from typing import Dict, TypedDict
>  
>  
>  class ConstTxtStyleClass(TypedDict):
> ```

Great find! Thank you. Do you want to create a PR or should I?

PR #562, opened 2023-09-05
558 improve capabilities handling

### LINKED ISSUE #558: Capabilities handling improvement
Capabilities handling needs improvement

When built without a capabilities file (this should not happen if the repository is properly cloned), some issues arise as reported in #556.

In order to address this:
* improve error message if no capabilities file is found
* make sure that if no capabilities file is found a bare minimum is generated (and thus no empty structure is accessed as in #556).
* improve the documentation for builds from git (put a section in the developer documentation and link from section installation)

patkan:
### Description
Improves the handling and documentation in case of missing capabilities file.
fixes #558

### Tested with
Tested with unit test.

### DIFF
--- a/doc/user/installation.rst
+++ b/doc/user/installation.rst
@@ -29,6 +29,20 @@ The following installation options exist:
  * `win32`: install packages required for win32 printing (only Windows)
  * `cups`: install packages required for CUPS printing
 
+Other installation methods
+--------------------------
+Officially, no other installation methods are supplied.
+
+If you want to install nevertheless from another source,
+please make sure that you have received the correct package
+and that the capabilities data is properly integrated.
+When packaging from source please read the developer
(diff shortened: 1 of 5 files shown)


doc/user/installation.rst
```diff
@@ -29,6 +29,20 @@ The following installation options exist:
...
+--------------------------
+Officially, no other installation methods are supplied.
+
+If you want to install nevertheless from another source,
+please make sure that you have received the correct package
+and that the capabilities data is properly integrated.
+When packaging from source please read the developer
+information in :ref:`developer-manual-repository`.
+
+If your packaging methode breaks the resource system from setuptools,
```
patkan, line 42:
```suggestion
If your packaging method breaks the resource system from setuptools,
```

PR #591, opened 2023-10-30
Restore types-pywin32

belono:
### Description
As mentioned in https://github.com/python-escpos/python-escpos/pull/554#issuecomment-1759762685 the types-pywin32 package was causing False positives  to the type checking.
I reported this issue to the python/typeshed proyect https://github.com/python/typeshed/issues/10891 where it was fixed and finally yesterday the types-pywin32 pypi repo was updated to v.306.0.0.6 that includes this fix.

We can restore the usage of the types-pywin32 dependency to enhance the type checking of the Win32Raw printer.

### Tested with

### DIFF
--- a/src/escpos/printer/win32raw.py
+++ b/src/escpos/printer/win32raw.py
@@ -148,7 +148,7 @@ class Win32Raw(Escpos):
         win32print.ClosePrinter(self._device)
         self._device = False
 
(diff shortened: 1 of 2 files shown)


patkan:
Thank you for reporting upstream!

PR #607, opened 2023-12-10
Python3-six-removal

a-detiste:
I never got my device to work ... sadness,
I'm better at software that hardware.

I gifted identical device to a hackerlab and while googling for support I found this one back https://iooner.io/epson-vfd-hack/ :-) :-) 

So anyway the weird mixture of Six & Typing ticked me and there I can help.

https://wiki.debian.org/Python3-six-removal

But ... ! While removing six I got scared and now I'd prefer to cleanup the annotations first.

(I can also package this for Debian)

### DIFF
--- a/src/escpos/escpos.py
+++ b/src/escpos/escpos.py
@@ -9,7 +9,7 @@ This module contains the abstract base class :py:class:`Escpos`.
 :copyright: Copyright (c) 2012-2017 Bashlinux and python-escpos
 :license: MIT
 """
-
+from __future__ import annotations
 
 import textwrap
 import warnings
--- a/src/escpos/escpos.py
+++ b/src/escpos/escpos.py
@@ -16,7 +16,7 @@ import warnings
 from abc import ABCMeta, abstractmethod  # abstract base class support
 from re import match as re_match
 from types import TracebackType
-from typing import Any, List, Literal, Optional, Union
+from typing import Any, List, Literal, Optional, Type, Union
 
 import barcode
 import qrcode
@@ -1438,7 +1438,7 @@ class EscposIO:
         return self
 
     def __exit__(
-        self, type: type[BaseException], value: BaseException, traceback: TracebackType
+        self, type: Type[BaseException], value: BaseException, traceback: TracebackType
     ) -> None:
         """Cut and close if configured.
 
--- a/src/escpos/escpos.py
+++ b/src/escpos/escpos.py
@@ -16,7 +16,7 @@ import warnings
 from abc import ABCMeta, abstractmethod  # abstract base class support
 from re import match as re_match
 from types import TracebackType
-from typing import Any, List, Literal, Optional, Union
+from typing import Any, Literal, Optional, Union
 
 import barcode
 import qrcode
@@ -1268,10 +1268,10 @@ class Escpos(object, metaclass=ABCMeta):
         else:
             self._raw(PANEL_BUTTON_OFF)
 
-    def query_status(self, mode: bytes) -> List[int]:
+    def query_status(self, mode: bytes) -> bytes:
         """Query the printer for its status.
 
-        Returns an array of integers containing it.
+        Returns byte array containing it.
 
         :param mode: Integer that sets the status mode queried to the printer.
             - RT_STATUS_ONLINE: Printer status.
--- a/src/escpos/escpos.py
+++ b/src/escpos/escpos.py
@@ -866,8 +866,7 @@ class Escpos(object):
         :param txt: text to be printed
         :raises: :py:exc:`~escpos.exceptions.TextError`
         """
-        txt = six.text_type(txt)
-        self.magic.write(txt)
+        self.magic.write(str(txt))
 
     def textln(self, txt=""):
         """Print alpha-numeric text with a newline.
@@ -1420,7 +1419,7 @@ class EscposIO(object):
         params = dict(self.params)
         params.update(kwargs)
 
-        if isinstance(text, six.text_type):
+        if isinstance(text, str):
             lines = text.split("\n")
         elif isinstance(text, list) or isinstance(text, tuple):
             lines = text
@@ -1433,10 +1432,7 @@ class EscposIO(object):
         # TODO flush? or on print? (this should prob rather be handled by the _raw-method)
         for line in lines:
             self.printer.set(**params)
-            if isinstance(text, six.text_type):
-                self.printer.text("{0}\n".format(line))
-            else:
-                self.printer.text("{0}\n".format(line))
+            self.printer.text("{0}\n".format(line))
 
     def close(self):
         """Close printer.
--- a/src/escpos/escpos.py
+++ b/src/escpos/escpos.py
@@ -1131,7 +1131,7 @@ class Escpos(object, metaclass=ABCMeta):
             try:
                 self._raw(CD_KICK_DEC_SEQUENCE(*pin))
             except TypeError as err:
-                raise CashDrawerError(err)
+                raise CashDrawerError(str(err))
 
     def linedisplay_select(self, select_display: bool = False) -> None:
         """Select the line display or the printer.
--- a/src/escpos/escpos.py
+++ b/src/escpos/escpos.py
@@ -154,6 +154,10 @@ class Escpos(object, metaclass=ABCMeta):
         """Open a printer device/connection."""
         pass
 
+    def close(self):
+        """Close a printer device/connection."""
+        pass
+
     @abstractmethod
     def _raw(self, msg: bytes) -> None:
         """Send raw data to the printer.
--- a/src/escpos/escpos.py
+++ b/src/escpos/escpos.py
@@ -15,7 +15,8 @@ import textwrap
 import warnings
 from abc import ABCMeta, abstractmethod  # abstract base class support
 from re import match as re_match
-from typing import List, Literal, Optional, Union
+from types import TracebackType
+from typing import List, Literal, Optional, Union, Any
 
 import barcode
 import qrcode
@@ -107,8 +108,7 @@ SW_BARCODE_NAMES = {
 }
 
 
-@six.add_metaclass(ABCMeta)
-class Escpos(object):
+class Escpos(object, metaclass=ABCMeta):
     """ESC/POS Printer object.
 
     This class is the abstract base class for an Esc/Pos-printer. The printer implementations are children of this
@@ -164,7 +164,7 @@ class Escpos(object):
         """
         pass
 
-    def _read(self):
+    def _read(self) -> bytes:
         """Read from printer.
 
         Returns a NotImplementedError if the instance of the class doesn't override this method.
@@ -250,7 +250,7 @@ class Escpos(object):
             header = (
                 GS
                 + b"v0"
-                + six.int2byte(density_byte)
+                + bytes((density_byte,))
                 + self._int_low_high(im.width_bytes, 2)
                 + self._int_low_high(im.height, 2)
             )
@@ -263,8 +263,8 @@ class Escpos(object):
             )
             tone = b"0"
             colors = b"1"
-            ym = six.int2byte(1 if high_density_vertical else 2)
-            xm = six.int2byte(1 if high_density_horizontal else 2)
+            ym = b"1" if high_density_vertical else b"2"
+            xm = b"1" if high_density_horizontal else b"2"
             header = tone + xm + ym + colors + img_header
             raster_data = im.to_raster_format()
             self._image_send_graphics_data(b"0", b"p", header + raster_data)
@@ -857,7 +857,7 @@ class Escpos(object):
         image = my_code.writer._image
         self.image(image, impl=impl, center=center)
 
-    def text(self, txt):
+    def text(self, txt:str) -> None:
         """Print alpha-numeric text.
 
         The text has to be encoded in the currently selected codepage.
@@ -868,7 +868,7 @@ class Escpos(object):
         """
         self.magic.write(str(txt))
 
-    def textln(self, txt=""):
+    def textln(self, txt: str = "") -> None:
         """Print alpha-numeric text with a newline.
 
         The text has to be encoded in the currently selected codepage.
@@ -879,7 +879,7 @@ class Escpos(object):
         """
         self.text("{}\n".format(txt))
 
-    def ln(self, count=1):
+    def ln(self, count: int = 1) -> None:
         """Print a newline or more.
 
         :param count: number of newlines to print
@@ -890,7 +890,7 @@ class Escpos(object):
         if count > 0:
             self.text("\n" * count)
 
-    def block_text(self, txt, font="0", columns=None):
+    def block_text(self, txt, font="0", columns=None) -> None:
         """Print text wrapped to specific columns.
 
         Text has to be encoded in unicode.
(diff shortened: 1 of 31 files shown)


a-detiste:
Ok there's something wrong with the CI itself

src/escpos/escpos.py
```diff
@@ -1433,23 +1434,22 @@ def writelines(self, text, **kwargs):
         # TODO flush? or on print? (this should prob rather be handled by the _raw-method)
         for line in lines:
             self.printer.set(**params)
-            if isinstance(text, six.text_type):
-                self.printer.text("{0}\n".format(line))
-            else:
-                self.printer.text("{0}\n".format(line))
+            self.printer.text("{0}\n".format(line))
```
patkan, line 1437:
```suggestion
            self.printer.text(f"{line}\n")
```

src/escpos/escpos.py
```diff
@@ -263,8 +263,8 @@ def image(
             )
             tone = b"0"
             colors = b"1"
-            ym = six.int2byte(1 if high_density_vertical else 2)
-            xm = six.int2byte(1 if high_density_horizontal else 2)
+            ym = b"1" if high_density_vertical else b"2"
+            xm = b"1" if high_density_horizontal else b"2"
```
patkan, line 267:
```suggestion
            xm = b"\x01" if high_density_horizontal else b"\x02"
```
Shouldn't that be hex? (Otherwise I guess you are using the byte representation of ASCII "1"

src/escpos/printer/network.py
```diff
@@ -110,16 +110,18 @@ def open(self, raise_not_found: bool = True) -> None:
...
         logging.info("Network printer enabled")
 
-    def _raw(self, msg):
+    def _raw(self, msg: bytes) -> None:
         """Print any command sent in raw format.
 
         :param msg: arbitrary code to be printed
         :type msg: bytes
         """
+        assert self.device
```
patkan, line 119: What is the motivation behind this assertion? If I remember correctly, the device property should open the connection automatically on access.

src/escpos/printer/network.py
```diff
@@ -110,16 +110,18 @@ def open(self, raise_not_found: bool = True) -> None:
...
         self.device.sendall(msg)
 
-    def _read(self):
+    def _read(self) -> bytes:
         """Read data from the TCP socket."""
+        assert self.device
```
patkan, line 124: Same here as in line 119

patkan (commented):
Thank you for the contribution.
First review pass, I will check again with more time. Please have a look at the comments.
As far as I know the type annotation for a None return type is technically not necessary, but doesn't do any harm either, so why not.

It would be great if you would package for Debian, maybe we should discuss in a discussion what would be needed in order to do it from this repository (on a separate branch?)

src/escpos/escpos.py
a-detiste: I'm not used to these new kind of f-string.

src/escpos/escpos.py
a-detiste: WHOOPS !

src/escpos/printer/network.py
a-detiste: I use `mypy --strict`to lint my code; mypy can't know that self.device has been initialized somehow in the lifecycle of the object, it's the only way to tell it `yes it's fine`

src/escpos/escpos.py
patkan: In the end it is just a more compact form of the format syntax (without the boilerplate like this: `"{thing}".format(thing=thing)`)

src/escpos/printer/network.py
patkan: OK, agreed, that is a bit much to expect from a static code checker :-)

a-detiste:
> what would be needed in order to do it from this repository (on a separate branch?)

this is against the common practices ... this way you would get a .deb file not a Debian package.

Debian is still obsessed about .tar.gz and these would then be loaded on https://salsa.debian.org/
(to create another .tar.gz for the build servers :-) )

patkan:
> > what would be needed in order to do it from this repository (on a separate branch?)
> 
> this is against the common practices ... this way you would get a .deb file not a Debian package.
> 
> Debian is still obsessed about .tar.gz and these would then be loaded on https://salsa.debian.org/ (to create another .tar.gz for the build servers :-) )

OK, we can do whatever is best practice. The only thing that would be important to me is that we find a way that could be "recoverable" if you would stop packaging python-escpos. I have tried (with the organization) to have the python-escpos package at least somehow hosted in a such a way that somebody else could step up and take this over with (hopefully) not too much trouble, should I stop doing this some day. (Ideally, there would be a transition, but you never know)

a-detiste:
I'm aware of this too strong ownership of packages in the Debian project.

I got shouted at by some survivalist living of the grid 11 months in a row
and going online once a year. "It's my package"

I will package it under the "Python team" with a disclaimer that anyone else can update it.

patkan:
> I'm aware of this too strong ownership of packages in the Debian project.
> 
> I got shouted at by some survivalist living of the grid 11 months in a row and going online once a year. "It's my package"
> 
> I will package it under the "Python team" with a disclaimer that anyone else can update it.

That would be great :-)



I have created a PR that refactors everything to f-strings: #608. If you do not mind merging or rebasing your branch, I would merge that into master.

a-detiste:
>  If you do not mind merging

I do not use an ide ... still nano ... :-|

patkan:
> > If you do not mind merging
> 
> I do not use an ide ... still nano ... :-|

I have checked and it is rather easy to merge. If that is OK for you, I would merge #608 and merge the changes into your branch, then you just have to pull :-)

a-detiste:
Yes please.

patkan:
> Yes please.

Done

patkan:
Now there is only one thing left in the CI:

```
  py38: commands[0]> pytest
  ImportError while loading conftest '/home/runner/work/python-escpos/python-escpos/test/conftest.py'.
  test/conftest.py:4: in <module>
      from escpos.printer import LP, CupsPrinter, Dummy, File, Network, Serial, Usb, Win32Raw
  .tox/py38/lib/python3.8/site-packages/escpos/printer/__init__.py:4: in <module>
      from .cups import CupsPrinter
  .tox/py38/lib/python3.8/site-packages/escpos/printer/cups.py:16: in <module>
      from ..escpos import Escpos
  .tox/py38/lib/python3.8/site-packages/escpos/escpos.py:1363: in <module>
      class EscposIO:
  .tox/py38/lib/python3.8/site-packages/escpos/escpos.py:1441: in EscposIO
      self, type: type[BaseException], value: BaseException, traceback: TracebackType
  E   TypeError: 'type' object is not subscriptable
```

I will soon be offline for today and cannot approve the CI runs. But you can also execute the tests locally with e.g. `tox -e py310` if you want to try out changes.

patkan:
I think the issue was that one: 369e709fc29d82b0d7d92009dc9c6f37c1190e65, feel free to revert if that is not what you meant.

a-detiste:
I would have proposed a `from __future__ import annotations`

a-detiste:
Your solution is the perfect one for python3.8, which I'm not used to work with (I do 2.6, 2.7, 3.7, 3.11, 3.12)

patkan:
> Your solution is the perfect one for python3.8, which I'm not used to work with (I do 2.6, 2.7, 3.7, 3.11, 3.12)

OK, but then I would prefer that we git revert 369e709fc29d82b0d7d92009dc9c6f37c1190e65, because I would prefer more the way how it will be done in the future.

patkan:
> > Your solution is the perfect one for python3.8, which I'm not used to work with (I do 2.6, 2.7, 3.7, 3.11, 3.12)
> 
> OK, but then I would prefer that we git revert [369e709](https://github.com/python-escpos/python-escpos/commit/369e709fc29d82b0d7d92009dc9c6f37c1190e65), because I would prefer more the way how it will be done in the future.

I have cleaned up my commit

a-detiste:
> I would prefer more the way how it will be done in the future.

pyupgrade can do it automaticaly for you when you bump your baseline.

`Optional[a]` becomes `a | None`
`Union[a|b]` becomes `a | b`

most uppercase imports from `typing` are replaced by lowercase ones from standard library
`List` -> `list`

but it's easier to keep it 3.8 coherant for now.

a-detiste:
> src/escpos/escpos.py:1278: error: Incompatible return value type (got "bytes", expected "list[int]")  [return-value]

this looks like a real bug

patkan:
> > src/escpos/escpos.py:1278: error: Incompatible return value type (got "bytes", expected "list[int]")  [return-value]
> 
> this looks like a real bug

Is this about the return value of the cups driver? If yes, maybe @belono can help? It looks like the return value of _read is not compatible with the rest of the lib for this driver.

a-detiste:
It could works thanks to duck typing: in some context bytes can be itterrated as single int s. But it would mire readable to stick with one data type for this object interface.

belono:
> > > src/escpos/escpos.py:1278: error: Incompatible return value type (got "bytes", expected "list[int]")  [return-value]
> > 
> > 
> > this looks like a real bug
> 
> Is this about the return value of the cups driver? If yes, maybe @belono can help? It looks like the return value of _read is not compatible with the rest of the lib for this driver.

Should the cups driver's `_read()` method return bytes? If so, would this return values be correct?
`b"0"` for the 'idle' state
`b"8"` for 'printing a job' or 'stopped' states

a-detiste:
If it's really about bytes emulated/encapsulated by cups it's better to use bytes all the way.

a-detiste:
Ok I'm further.

`error: "Escpos" has no attribute "close" ` -> indeed the destructor will call in cascade a method that doesn't exist

patkan:
> Ok I'm further.
> 
> `error: "Escpos" has no attribute "close" ` -> indeed the destructor will call in cascade a method that doesn't exist

The method should be implemented by the printer drivers. Is this not the case?

a-detiste:
There's Escpos && EscposIO:

One close() is missing

```
class Escpos

    def __del__(self):
        """Call self.close upon deletion."""
        self.close()


class EscposIO:
    r"""ESC/POS Printer IO object.
       def __init__(
        self, printer: Escpos, autocut: bool = True, autoclose: bool = True, **kwargs
    ) -> None:
        """Initialize object.
        :param printer: An EscPos-printer object
        """
        self.printer = printer

 def close(self) -> None:
        """Close printer.

        Called upon closing the `with`-statement.
        """
        self.printer.close()
```

a-detiste:
It's a baseclass, so maybe just addig "def close(): pass" is enough

a-detiste:
I know nothing about win32, but the typing in the stubs _feels_ wrong.

```
/usr/lib/python3/dist-packages/win32-stubs/win32print.pyi:def WritePrinter(hprinter: _win32typing.PyPrinterHANDLE, buf: str): ...
```

a-detiste:
BTW the `) -> None:` has one specific purpose: it tells mypy that annotation job has been done and it can proceed with further checking. This makes more hints show up.

patkan:
Hi @a-detiste, are you ready with the change? (CI is passing)
Would it be possible to change the mypy ci step to be more strict? (Or do we have remaining issues there?)

a-detiste:
Hi, I would like to slow download before my Holydays (from 20), so let's wrap this up.

mypy --strict generates 400 errors, it's wat too early.

patkan:
Ok, well this would be quite a lot still, haven't checked mypy directly.

Should we merge this pr now? Then you can open a new one for the next steps when you get to it

a-detiste:
Yes please merge.

I will track on my side the dependecies that needs:
- annotations
- packaging in Debian

a-detiste:
I need to package python3-barcode first

a-detiste:
please include examples/ in mypy run.
this helped me understanding & fixing this previous typing error

-tox.ini:commands = mypy src test
+tox.ini:commands = mypy src test examples

```
-    def __init__(
-        idVendor: str = "",
-        idProduct: str = "",
-        usb_args: Dict[str, str] = {},
    
+    def __init__(
+        self,
+        idVendor: Optional[int] = None,
+     idProduct: Optional[int] = None,
+     usb_args: Dict[str, Union[str, int]] = {},
```

patkan (approved):
For me these changes would be ok. @belono what do you think?

belono:
I left a comment in the [cups.py](https://github.com/python-escpos/python-escpos/pull/607/files#diff-b7fcda07446fce62bb555abb5722fbc061848fdf0fbc2744bd38e122d4935ee4R208) and another one in [test_printer_cups.py](https://github.com/python-escpos/python-escpos/pull/607/files/45829b8ecb9d6ec2f33336da5b572801a49be233..254eae07ef345bef788e108176b659c2353e8449#diff-8fba37f83986b38b6d5a02c153258f0511aff1e2163c72663397c2b983686e5aR176) some days ago. :smile: 

Please, fix that before merge.

patkan:
> I left a comment in the [cups.py](https://github.com/python-escpos/python-escpos/pull/607/files#diff-b7fcda07446fce62bb555abb5722fbc061848fdf0fbc2744bd38e122d4935ee4R208) and another one in [test_printer_cups.py](https://github.com/python-escpos/python-escpos/pull/607/files/45829b8ecb9d6ec2f33336da5b572801a49be233..254eae07ef345bef788e108176b659c2353e8449#diff-8fba37f83986b38b6d5a02c153258f0511aff1e2163c72663397c2b983686e5aR176) some days ago. 😄
> 
> Please, fix that before merge.

Are you sure that you have sent the review? Could it be still pending? I cannot see those comments unfortunately. :confused:

belono (changes requested):
Please, modify the `CupsPrinter._read()` method and its test.

src/escpos/printer/cups.py
```diff
@@ -199,16 +199,16 @@ def _clear(self) -> None:
...
 
-    def _read(self) -> List[int]:
+    def _read(self) -> bytes:
         """Return a single-item array with the accepting state of the print queue.
 
         states: idle = [3], printing a job = [4], stopped = [5]
         """
         printer = self.printers.get(self.printer_name, {})
         state = printer.get("printer-state")
         if not state:
```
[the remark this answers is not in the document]
a-detiste: I think a lot of constants.py should be rewriten using Enum / IntEnum instead of having magic numbers that needs commenting here and there... later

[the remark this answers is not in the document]
patkan: you could use `RT_MASK_ONLINE` in this case.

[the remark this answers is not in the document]
belono: The reason I didn't suggest RT_MASK_ONLINE is that its name is not consistent with the meaning of the returning value. According to [Esc/Pos](https://download4.epson.biz/sec_pubs/pos/reference_en/escpos/dle_eot.html), the int value 8 is for **Offline** status.

We could modify that constant and add a new one for offline:
```python
RT_MASK_ONLINE: int = 0
RT_MASK_OFFLINE: int = 8
```
Then we could do:
```python
if not state or state in [4, 5]:
    return RT_MASK_OFFLINE  # offline
return RT_MASK_ONLINE  # online
```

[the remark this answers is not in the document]
patkan: I have created an issue so that we do not forget and introduced your first example, @belono so that we have a solution for now.

PR #624, opened 2024-01-24
Add sleep in sending fragments

aerialist:
Adding sleep prevents "USBTimeoutError: [Errno 110] Operation timed out".

### Tested with
Raspberry Pi with this printer via USB
https://a.aliexpress.com/_oFusnk4


belono (changes requested):
After fixing the import sorting it will be OK to me.

PR #645, opened 2024-06-02
New feature: Software columns

### LINKED ISSUE #27: align left and right on the same line
```
What steps will reproduce the problem?
1. use Epson TM-T20
2. execute the following code
3.

What is the expected output? What do you see instead?
I would like to format a line to begin on the left with some words ("article" 
in my example) and then align on the right to write the price (15.00 € here)

What version of the product are you using? On what operating system?
python-escpos 1.0-1 on ubuntu 13.10

Please provide any additional information below.

code :
Epson.set("LEFT", "A", "", 1, 1)
Epson.text("Article 1")
Epson.set("RIGHT", "A", "", 1, 1)
Epson.text("15.00 \xd5")
Epson.text("\n")

```

Original issue reported on code.google.com by `shu...@gmail.com` on 15 May 2014 at 10:29

### LINKED ISSUE #635: Tabbed text
I have:

Cannot get tabbed text. It stays on the left I am trying achieve item left price right. But I wanted to see how the tab works first but I can't get it to go the the right. Any pointers would be highly appreciated.
Thank you

```
printer = Win32Raw()
printer.open()
printer.control("HT", count=8)
printer.text("tabbed text 8")
```
This prints left aligned - what am I missing?
### Device info
Printer is Aures ODP 333 (it supports esc/pos HT command)

 **python-escpos version:** 3.1

 **python version:** 3.12.1

 **operating system:** Windows 11

belono:
### Description

This implements `software_columns()` as a new feature to help in the tricky task of printing texts in columns or aligning multiple texts in different alignments within the same line.

The method receives a list of strings and formats them as horizontal columns by adding spaces between each element. If some text is too long it is wrapped to the column width by adding new lines and even truncating words that are too long by adding a placeholder.
It is also possible to set the width and the text alignment of each column independently. The method is flexible enough to repeat the last width and/or alignment till the last element if there are more text elements than column widths or alignments.
- Example 1: The user passes a list of 4 text items and wants the first one to be 20 characters wide and 10 for the rest, then he/she only needs to pass the value `widths=[20, 10]` to the method so it will be expanded to '[20, 10, 10, 10]`.
- Example 2: The user passes a list of 4 text items and wants all them to be the same size along a maximum total width of 40 characters, then he/she simply passes the value `widths=40` so it will be expanded to `[10, 10, 10, 10]`.

This is also true for the alignment parameter:
- On a list of 4 text items `align=['left', 'center', 'right']` would expand to `['left', 'center', 'right', 'right']`. Alternatively a single value `align='center'` would expand to `['center', 'center', 'center', 'center']`.

At the moment, the user is responsible for not exceeding the width of the paper when passing the list of `widths` parameter, but this behavior could be modified by adding a check, an exception or whatever.

As always all suggestions, corrections, etc. are very welcome.

Closes #27
Closes #635

### Tested with
- Epson TM-U210

### DIFF
--- a/src/escpos/escpos.py
+++ b/src/escpos/escpos.py
@@ -107,6 +107,8 @@ SW_BARCODE_NAMES = {
     for name in barcode.PROVIDED_BARCODES
 }
 
+Alignment = Union[Literal["center", "left", "right"], str]
+
 
 class Escpos(object, metaclass=ABCMeta):
     """ESC/POS Printer object.
--- a/src/escpos/escpos.py
+++ b/src/escpos/escpos.py
@@ -899,6 +899,26 @@ class Escpos(object, metaclass=ABCMeta):
         col_count = self.profile.get_columns(font) if columns is None else columns
         self.text(textwrap.fill(txt, col_count))
 
+    @staticmethod
+    def padding(
+        text: str,
+        width: int,
+        align: Alignment = "center",
+    ) -> str:
+        """Add fill space to meet the width.
+
+        The align parameter sets the alignment of the text in space.
+        """
+        align = align.lower()
+        if align == "center":
+            text = f"{text:^{width}}"
+        elif align == "left":
+            text = f"{text:<{width}}"
+        elif align == "right":
+            text = f"{text:>{width}}"
+
+        return text
+
     def set(
         self,
         align: Optional[str] = None,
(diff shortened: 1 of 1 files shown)


patkan (approved):
Thank you! Looks good to me and the CI hasn't found issues.

PR #651, opened 2024-07-19
New feature: Software columns - Part 3: Tests

belono:
### Description
This is part 3 of the series on implementing [software columns](https://github.com/python-escpos/python-escpos/pull/645).

This adds some simple tests to the battery of tests.

### Tested with
My local pytest suite


patkan (approved):
Thank you! Nothing to add from my side :-)

lilkimo:
Cool feature! It would be nice if it was available in the PyPi package

PR #680, opened 2025-03-06
Feature/kanji commands

kymok:
### Description

This PR adds Kanji-related feature.
- [ESC/POS's Kanji feature](https://download4.epson.biz/sec_pubs/pos/reference_en/escpos/) is implemented.

### Tested with
- EPSON TM-m10 (Japanese model)
  - Supports only ISO-2022-JP and Shift-JIS encoding

IMG_3368

### Note
- Neither Korean, Traditional Chinese, Simplified Chinese nor Japanese (Shift-JIS 2004) are tested since I don't have any printer that accepts those encodings.


kymok:
Marking as draft until being able to build docs locally

examples/kanji.py
```diff
@@ -0,0 +1,53 @@
...
+p.set_kanji_decoration()
+p.ln()
+
+p.kanji_text("こんにちは世界！\n")
+p.ln()
+
+p.define_user_defined_kanji(b"\x77\x7e", checkerboard_kanji)
+p._enter_kanji_mode()
+p._raw(b"\x77\x7e")
+p._exit_kanji_mode()
```
patkan, line 50: Do I understand this feature correctly, this sequence is used to write the user defined kanji? In this case I would propose an API function such as `write_user_defined_kanji(string)` in order to write the kanji and prevent forcing users to use the "internal" functions.

src/escpos/escpos.py
```diff
@@ -1494,6 +1507,204 @@ def buzzer(self, times: int = 2, duration: int = 4) -> None:
...
+            # ISO-2022-JP encoding is a stateful encoding.
+            # We need to enter and exit Kanji mode.
+            self._iso2022jp_text(text)
+        else:
+            encoded_text = text.encode(self.kanji_encoding, "ignore")
+            self._enter_kanji_mode()
+            self._raw(encoded_text)
+            self._exit_kanji_mode()
+
+    def _iso2022jp_text(self, text: str) -> None:
```
patkan, line 1537: For now this could be kept in escpos.py, but if this functionality grows more, we should think about a separate module (like magicencode)

src/escpos/escpos.py
```diff
@@ -1494,6 +1507,204 @@ def buzzer(self, times: int = 2, duration: int = 4) -> None:
...
+            "big5",  # FIXME test with real device,
+            "gb2312",  # FIXME test with real device,
+            "gb18030",  # FIXME test with real device,
+        ],
+    ):
+        """Select the Kanji encoding.
+
+        This command is available only for Japanese model printers.
+
+        :param code: Encoding.
```
patkan, line 1646:
```suggestion

        .. todo:: Test the encodings marked above with `FIXME` with a real device.
```

Please add a short todo for the FIXME above so that it shows up in this list: https://python-escpos--680.org.readthedocs.build/en/680/dev/todo.html

src/escpos/escpos.py
```diff
@@ -1494,6 +1507,204 @@ def buzzer(self, times: int = 2, duration: int = 4) -> None:
 
         self._raw(BUZZER + six.int2byte(times) + six.int2byte(duration))
 
+    def _enter_kanji_mode(self):
+        """Enter Kanji mode."""
```
patkan, line 1511:
```suggestion
    def _enter_kanji_mode(self) -> None:
```
Please annotate a return type None for completeness.

src/escpos/escpos.py
```diff
@@ -1494,6 +1507,204 @@ def buzzer(self, times: int = 2, duration: int = 4) -> None:
...
+        self._raw(KANJI_ENTER_KANJI_MODE)
+
+    def _exit_kanji_mode(self):
+        """Exit Kanji mode."""
```
patkan, line 1515:
```suggestion
    def _exit_kanji_mode(self) -> None:
```
Please annotate a return type None for completeness.

src/escpos/escpos.py
patkan:
```suggestion
    def _iso2022jp_text(self, text: str) -> None:
```

Please annotate a return type None for completeness.

src/escpos/escpos.py
```diff
@@ -1494,6 +1507,204 @@ def buzzer(self, times: int = 2, duration: int = 4) -> None:
...
+                )
+
+    def set_kanji_decoration(
+        self,
+        *,
+        double_width: bool = False,
+        double_height: bool = False,
+        underline: Literal[0, 1, 2] = 0,
+    ):
+        """Set the Kanji print mode.
```
patkan, line 1578:
```suggestion
    ) -> None:
```
Please annotate a return type None for completeness.

src/escpos/escpos.py
```diff
@@ -1494,6 +1507,204 @@ def buzzer(self, times: int = 2, duration: int = 4) -> None:
...
+        if double_height:
+            n |= 0x08
+        self._raw(KANJI_PRINT_MODE + six.int2byte(n))
+        self.set_kanji_underline(underline)
+
+    def set_kanji_underline(
+        self,
+        underline: Literal[0, 1, 2] = 0,
+    ):
+        """Set the Kanji underline mode.
```
patkan, line 1596:
```suggestion
    ) -> None:
```
Please annotate a return type None for completeness.

src/escpos/escpos.py
```diff
@@ -1494,6 +1507,204 @@ def buzzer(self, times: int = 2, duration: int = 4) -> None:
...
+            2 = Set underline with 2 dot width.
+        """
+        self._raw(KANJI_UNDERLINE + six.int2byte(underline))
+
+    def define_user_defined_kanji(
+        self,
+        code: bytes,
+        data: bytes,
+    ):
+        """Set a user defined Kanji character.
```
patkan, line 1613:
```suggestion
    ) -> None:
```
Please annotate a return type None for completeness.

src/escpos/escpos.py
```diff
@@ -1494,6 +1507,204 @@ def buzzer(self, times: int = 2, duration: int = 4) -> None:
...
+        :param code: The Kanji code.
+        :param data: The Kanji data.
+        """
+        self._raw(KANJI_DEFINE_USER_DEFINED + code + data)
+
+    def delete_user_defined_kanji(
+        self,
+        code: bytes,
+    ):
+        """Delete a user defined Kanji character.
```
patkan, line 1624:
```suggestion
    ) -> None:
```
Please annotate a return type None for completeness.

src/escpos/escpos.py
```diff
@@ -1494,6 +1507,204 @@ def buzzer(self, times: int = 2, duration: int = 4) -> None:
...
+            "iso2022_jp",
+            "shift_jis",
+            "shift_jis_2004",
+            "euc_kr",  # FIXME test with real device,
+            "big5",  # FIXME test with real device,
+            "gb2312",  # FIXME test with real device,
+            "gb18030",  # FIXME test with real device,
+        ],
+    ):
+        """Select the Kanji encoding.
```
patkan, line 1642: Please annotate a return type None for completeness.

src/escpos/escpos.py
```diff
@@ -1494,6 +1507,204 @@ def buzzer(self, times: int = 2, duration: int = 4) -> None:
...
+            self.kanji_encoding = encoding
+        else:
+            raise ValueError("Invalid encoding")
+
+    def set_kanji_spacing(
+        self,
+        left_spacing: int,
+        right_spacing: int,
+    ):
+        """Set the Kanji spacing.
```
patkan, line 1673:
```suggestion
    ) -> None:
```
Please annotate a return type None for completeness.

src/escpos/escpos.py
```diff
@@ -1494,6 +1507,204 @@ def buzzer(self, times: int = 2, duration: int = 4) -> None:
...
+        """
+        self._raw(
+            KANJI_SET_SPACING + six.int2byte(left_spacing) + six.int2byte(right_spacing)
+        )
+
+    def set_kanji_quadruple_size(
+        self,
+        enable: bool,
+    ):
+        """Set the Kanji quadruple size.
```
patkan, line 1688:
```suggestion
    ) -> None:
```
Please annotate a return type None for completeness.

src/escpos/escpos.py
```diff
@@ -1494,6 +1507,204 @@ def buzzer(self, times: int = 2, duration: int = 4) -> None:
...
+
+        :param enable: Enable quadruple size.
+        """
+        self._raw(KANJI_SET_QUADRUPLE_SIZE + six.int2byte(int(enable)))
+
+    def set_kanji_font(
+        self,
+        font: Literal[0, 1, 2],
+    ):
+        """Set the Kanji font.
```
patkan, line 1698:
```suggestion
    ) -> None:
```
Please annotate a return type None for completeness.

test/test_functions/test_function_kanji.py
```diff
@@ -0,0 +1,183 @@
...
+    assert instance.output == KANJI_ENTER_KANJI_MODE
+
+
+def test_exit_kanji_mode() -> None:
+    """should exit kanji mode."""
+    instance = printer.Dummy()
+    instance._exit_kanji_mode()
+    assert instance.output == KANJI_EXIT_KANJI_MODE
+
+
```
patkan, line 58:
```suggestion

def test_kanji_text_without_encoding() -> None:
    """Test behavior when no encoding is set."""
    instance = printer.Dummy()
    with pytest.raises(ValueError):
        instance.kanji_text("Hello世界Hello")

```
Please test also the negative case when no encoding is set.

patkan (changes requested):
Thank you very much for this change request. This looks good to me and I will merge it, I just have some small remarks. Please go through them and accept or comment depending on if you think it fits.

examples/kanji.py
kymok: Yes. That seems to be a better way to go. I'll implement them but the input will be hex value.

In some encoding system, some code points are left undefined, and in this example we're writing a new glyph to one of those code point. The undefined zone differs by both encoding and maybe printers, so basically user need to input hex value of code point manually.

patkan: I agree, a hex value is better for the code point (and aligns with the rest of the code that you wrote).

Issue #704, opened 2025-09-07
Migrate to uv


hasansezertasan:
"An extremely fast Python package and project manager, written in Rust.", says [uv documentation](https://docs.astral.sh/uv/).

"🔩 Includes a [pip-compatible interface](https://docs.astral.sh/uv/#the-pip-interface) for a performance boost with a familiar CLI."

PR #705, opened 2025-09-07
docs(examples): migrate to uv and PEP 723 inline script metadata

hasansezertasan:
Related to #704

## What

Migrate the runnable examples in `examples/` to [uv](https://docs.astral.sh/uv/) with [PEP 723](https://peps.python.org/pep-0723/) inline script metadata, so each example can be run directly with `uv run <example>.py` without a manual virtualenv or dependency install.

## Changes

- Add PEP 723 `# /// script` metadata blocks to the standalone examples (`barcodes.py`, `codepage_tables.py`, `qr_code.py`, `software_barcode.py`, `software_columns.py`, `weather.py`), declaring `requires-python`, dependencies, and a local editable `python-escpos` source.
- Tidy the example scripts (wrap top-level code in a `main()` entry point, minor cleanups) to fit the single-file script model.
- Replace the `examples/docker-flask/` example with `examples/flask-example/`, now managed by uv via `pyproject.toml` and `uv.lock`, with a uv-based `Dockerfile`.

## Notes

- No library/source changes — this only touches `examples/`.


examples/barcodes.py
```diff
@@ -1,11 +1,23 @@
+# /// script
+# requires-python = ">=3.12"
```
belono: We should require the minimum version of Python we currently support, otherwise a user would not be able to run this example script on a version lower than 3.12 while we currently support Python >=3.9.
```suggestion
# requires-python = ">=3.9"
```

examples/codepage_tables.py
```diff
@@ -1,9 +1,12 @@
+# /// script
+# requires-python = ">=3.12"
```
belono: Same on this file. We should require the minimum version of Python we currently support, otherwise a user would not be able to run this example script on a version lower than 3.12 while we currently support Python >=3.9.
```suggestion
# requires-python = ">=3.9"
```

examples/qr_code.py
```diff
@@ -1,21 +1,26 @@
+# /// script
+# requires-python = ">=3.12"
```
belono: The same here. We should require the minimum version of Python we currently support, otherwise a user would not be able to run this example script on a version lower than 3.12 while we currently support Python >=3.9.
```suggestion
# requires-python = ">=3.9"
```

examples/software_barcode.py
```diff
@@ -1,9 +1,21 @@
+# /// script
+# requires-python = ">=3.12"
```
belono: Here too. We should require the minimum version of Python we currently support, otherwise a user would not be able to run this example script on a version lower than 3.12 while we currently support Python >=3.9.
```suggestion
# requires-python = ">=3.9"
```

examples/software_columns.py
```diff
@@ -1,25 +1,38 @@
-""" Example for software_columns: Print text arranged into columns."""
+# /// script
+# requires-python = ">=3.12"
```
belono: And here. We should require the minimum version of Python we currently support, otherwise a user would not be able to run this example script on a version lower than 3.12 while we currently support Python >=3.9.
```suggestion
# requires-python = ">=3.9"
```

examples/weather.py
```diff
@@ -1,4 +1,9 @@
-#!/usr/bin/python
+# /// script
+# requires-python = ">=3.12"
```
belono: Same case. We should require the minimum version of Python we currently support, otherwise a user would not be able to run this example script on a version lower than 3.12 while we currently support Python >=3.9.
```suggestion
# requires-python = ">=3.9"
```

PR #707, opened 2025-09-11
Add Config.load_yaml_string()

Elias0419:
### Description

Add a new method Config.load_yaml_string() to allow loading configuration directly from a YAML string.

### DIFF
--- a/src/escpos/config.py
+++ b/src/escpos/config.py
@@ -31,7 +31,7 @@ class Config:
         self._has_loaded = False
         self._printer = None
 
-        self._printer_name = None
+        self._printer_name = ""
         self._printer_config = None
 
     def _reset_config(self) -> None:
@@ -42,8 +42,7 @@ class Config:
         """
         self._has_loaded = False
         self._printer = None
-
-        self._printer_name = None
+        self._printer_name = ""
         self._printer_config = None
 
     def _set_config(self, config) -> None:
@@ -51,11 +50,11 @@ class Config:
 
         :param config: config object loaded from file or string.
         """
-        self._printer_config = config.get("printer")
+        self._printer_config = config.get("printer", None)
         if self._printer_config is None:
             self._has_loaded = True
             return
-        printer_name = self._printer_config.pop("type")
+        printer_name = self._printer_config.get("type", "")
         class_names = {
             "usb": "Usb",
             "serial": "Serial",
--- a/src/escpos/config.py
+++ b/src/escpos/config.py
@@ -2,6 +2,7 @@
 
 This module contains the implementations of abstract base class :py:class:`Config`.
 """
+
 import os
 import pathlib
 
@@ -99,9 +100,9 @@ class Config:
             )
 
     def load_yaml_string(self, yaml_string) -> None:
-        """Load and parse a yaml configuration string.
+        """Load and parse a yaml configuration string or file-like object.
 
-        :param yaml_string: A string containing yaml formatted configuration
+        :param yaml_string: A string or file-like object containing yaml formatted configuration.
         """
         self._reset_config()
 
@@ -111,7 +112,6 @@ class Config:
             raise exceptions.ConfigSyntaxError("Error parsing YAML")
 
         self._set_config(config)
-
 
     def printer(self):
         """Return a printer that was defined in the config.
--- a/src/escpos/config.py
+++ b/src/escpos/config.py
@@ -45,6 +45,33 @@ class Config:
         self._printer_name = None
         self._printer_config = None
 
+    def _set_config(self, config) -> None:
+        """Set configuration variables.
+
+        :param config: config object loaded from file or string.
+        """
+        if "printer" in config:
+            self._printer_config = config["printer"]
+            printer_name = self._printer_config.pop("type")
+            class_names = {
+                "usb": "Usb",
+                "serial": "Serial",
+                "network": "Network",
+                "file": "File",
+                "dummy": "Dummy",
+                "cupsprinter": "CupsPrinter",
+                "lp": "LP",
+                "win32raw": "Win32Raw",
+            }
+            self._printer_name = class_names.get(printer_name.lower(), printer_name)
+
+            if not self._printer_name or not hasattr(printer, self._printer_name):
+                raise exceptions.ConfigSyntaxError(
+                    f'Printer type "{self._printer_name}" is invalid'
+                )
+
+        self._has_loaded = True
+
     def load(self, config_path=None):
         """Load and parse the configuration file using pyyaml.
 
@@ -74,27 +101,23 @@ class Config:
         except yaml.YAMLError:
             raise exceptions.ConfigSyntaxError("Error parsing YAML")
 
-        if "printer" in config:
-            self._printer_config = config["printer"]
-            printer_name = self._printer_config.pop("type")
-            class_names = {
-                "usb": "Usb",
-                "serial": "Serial",
-                "network": "Network",
-                "file": "File",
-                "dummy": "Dummy",
-                "cupsprinter": "CupsPrinter",
-                "lp": "LP",
-                "win32raw": "Win32Raw",
-            }
-            self._printer_name = class_names.get(printer_name.lower(), printer_name)
+        self._set_config(config)
 
-            if not self._printer_name or not hasattr(printer, self._printer_name):
-                raise exceptions.ConfigSyntaxError(
-                    f'Printer type "{self._printer_name}" is invalid'
-                )
 
-        self._has_loaded = True
+    def load_yaml_string(self, yaml_string):
+        """Load and parse a yaml configuration string.
+
+        :param yaml_string: A string containing yaml formatted configuration
+        """
+        self._reset_config()
+
+        try:
+            config = yaml.safe_load(yaml_string)
+        except yaml.YAMLError:
+            raise exceptions.ConfigSyntaxError("Error parsing YAML")
+
+        self._set_config(config)
+
 
     def printer(self):
         """Return a printer that was defined in the config.
--- a/src/escpos/config.py
+++ b/src/escpos/config.py
@@ -51,25 +51,26 @@ class Config:
 
         :param config: config object loaded from file or string.
         """
-        if "printer" in config:
-            self._printer_config = config["printer"]
-            printer_name = self._printer_config.pop("type")
-            class_names = {
-                "usb": "Usb",
-                "serial": "Serial",
-                "network": "Network",
-                "file": "File",
-                "dummy": "Dummy",
-                "cupsprinter": "CupsPrinter",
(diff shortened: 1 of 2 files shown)


belono:
Hi @Elias0419, and thanks for your contribution.

I like this feature and I'm glad you included tests with it.

The code looks good, although I'd like to make a few suggestions.

src/escpos/config.py
belono, line 1: Before this commit, `load` was responsible to clear the old configuration, parse the yaml string, and set the configuration. Now, this job is made by `load_yaml_string`, so we can remove it from `load` and delegate to `load_yaml_string`.

src/escpos/config.py
```diff
@@ -45,6 +45,33 @@ def _reset_config(self) -> None:
...
+            self._printer_name = class_names.get(printer_name.lower(), printer_name)
+
+            if not self._printer_name or not hasattr(printer, self._printer_name):
+                raise exceptions.ConfigSyntaxError(
+                    f'Printer type "{self._printer_name}" is invalid'
+                )
+
+        self._has_loaded = True
+
     def load(self, config_path=None):
```
belono:
```python
    def load(self, config_path=None):
        """Load and parse the configuration file using pyyaml.

        :param config_path: An optional file path, file handle, or byte string
            for the configuration file.
        """
        # self._reset_config()  # CHANGE: remove

        if not config_path:
            config_path = os.path.join(
                platformdirs.user_config_dir(self._app_name), self._config_file
            )
        if isinstance(config_path, pathlib.Path):
            # store string if posixpath
            config_path = config_path.as_posix()
        if not os.path.isfile(config_path):
            # supplied path is not a file --> assume default file
            config_path = os.path.join(config_path, self._config_file)
        try:
            with open(config_path, "rb") as config_file:
                # config = yaml.safe_load(config_file)  # CHANGE: remove
                self.load_yaml_string(config_file)  # CHANGE: call load_yaml_string()
        except EnvironmentError:
            raise exceptions.ConfigNotFoundError(
                f"Couldn't read config at {config_path}"
            )
        '''CHANGE: remove this block
        except yaml.YAMLError:
            raise exceptions.ConfigSyntaxError("Error parsing YAML")

        self._set_config(config)
        '''
```

src/escpos/config.py
```diff
@@ -45,6 +45,33 @@ def _reset_config(self) -> None:
...
+                "cupsprinter": "CupsPrinter",
+                "lp": "LP",
+                "win32raw": "Win32Raw",
+            }
+            self._printer_name = class_names.get(printer_name.lower(), printer_name)
+
+            if not self._printer_name or not hasattr(printer, self._printer_name):
+                raise exceptions.ConfigSyntaxError(
+                    f'Printer type "{self._printer_name}" is invalid'
+                )
```
belono: Mypy complains about this block: 'src/escpos/config.py:55: error: "None" has no attribute "pop"'
This would fix it:

```suggestion
        self._printer_config = config.get("printer")
        if not self._printer_config:
            return

        printer_name = self._printer_config.pop("type")
        class_names = {
            "usb": "Usb",
            "serial": "Serial",
            "network": "Network",
            "file": "File",
            "dummy": "Dummy",
            "cupsprinter": "CupsPrinter",
            "lp": "LP",
            "win32raw": "Win32Raw",
        }
        self._printer_name = class_names.get(printer_name.lower(), printer_name)

        if not self._printer_name or not hasattr(printer, self._printer_name):
            raise exceptions.ConfigSyntaxError(
                f'Printer type "{self._printer_name}" is invalid'
            )
```

src/escpos/config.py
belono: We could add the missing return type annotation:
```suggestion
    def load(self, config_path=None) -> None:
```

src/escpos/config.py
```diff
@@ -74,27 +101,23 @@ def load(self, config_path=None):
...
-            self._printer_name = class_names.get(printer_name.lower(), printer_name)
+        self._set_config(config)
 
-            if not self._printer_name or not hasattr(printer, self._printer_name):
-                raise exceptions.ConfigSyntaxError(
-                    f'Printer type "{self._printer_name}" is invalid'
-                )
 
-        self._has_loaded = True
+    def load_yaml_string(self, yaml_string):
```
belono: Again, we could add the missing return type annotation:
```suggestion
    def load_yaml_string(self, yaml_string) -> None:
```

test/test_config.py
belono, line 1: It would be great if you could also add the missing return type annotations for the functions in this file.

belono (commented):
I think that's all for now.

Feel free to discuss any of my suggestions if you disagree or if I haven't expressed my point well :ok_hand:

test/test_config.py
Elias0419: Done in [73f15bd](https://github.com/python-escpos/python-escpos/pull/707/commits/73f15bd91941cf0ccc3405851c7e99c4b48a985c)

src/escpos/config.py
Elias0419: Done in [bcba002](https://github.com/python-escpos/python-escpos/pull/707/commits/bcba002bfe1f47277d54dd2b5bc451b2adaf1d2a)

src/escpos/config.py
Elias0419: Done in [bcba002](https://github.com/python-escpos/python-escpos/pull/707/commits/bcba002bfe1f47277d54dd2b5bc451b2adaf1d2a)

src/escpos/config.py
Elias0419: Hi @belono. Thanks for having a look. I have a small nit to pick with this one: `load` currently sets  `_has_loaded` to True even when a printer is not found:

```
        if "printer" in config:
            self._printer_config = config["printer"]
         #...   
         
        self._has_loaded = True
```


We should set `_has_loaded` to True before the early return to maintain the same behavior:

```
    if not self._printer_config:
        self._has_loaded = True
        return
```

src/escpos/config.py
Elias0419: Done in [fe554a4](https://github.com/python-escpos/python-escpos/pull/707/commits/fe554a43f4ef68f1795afe949d66eb16d2e06e23)

src/escpos/config.py
Elias0419: Also if we use a generic falsy check here (`if not self._printer_config:`), then a config with an empty dict like `printer: {}` would be treated as "no printer section" and accepted.

In the original implementation this case actually raised a KeyError when trying to pop "type"

belono: I agree. The falsy check `if not self._printer_config:` is incorrect. We should check for `None`.

> then a config with an empty dict like printer: {} would be treated as "no printer section" and accepted

We really should add a test for this case. Something like this approach:
```python
from escpos import config

c = config.Config()
with pytest.raises(escpos.exceptions.ConfigSyntaxError):
    c.load_yaml_string("printer:\n   {}")
```

Elias0419: Currently, we'd need to check for the KeyError

```
    with pytest.raises(KeyError):
        c.load_yaml_string("printer: {}")
```
Unless you want to catch it and re-raise it as ConfigSyntaxError

```
if self._printer_config is None:
    self._has_loaded = True
    return
try:
    printer_name = self._printer_config.pop("type")
except KeyError:
    raise exceptions.ConfigSyntaxError('...')
```

belono: > We should set _has_loaded to True before the early return to maintain the same behavior:

After a second look at the old and the new code I realized that there is another way not involving early return and maintaining a similar behavior.

I'll made a code suggestion.

src/escpos/config.py
```diff
@@ -45,13 +46,40 @@ def _reset_config(self) -> None:
...
+    def _set_config(self, config) -> None:
+        """Set configuration variables.
+
+        :param config: config object loaded from file or string.
+        """
+        self._printer_config = config.get("printer")
+        if self._printer_config is None:
+            self._has_loaded = True
+            return
+        printer_name = self._printer_config.pop("type")
```
belono: This passes the tests battery.

I replaced `pop()` with `get()` because a I can't find a reason to remove the "type" key given that there is a check for the existence of a `printer_name` attribute in the printer class later. Anyway, I think that `pop()` could be restored cleanly if needed.

```suggestion
        self._printer_config = config.get("printer", None)
        printer_name = ""
        if self._printer_config:
            printer_name = self._printer_config.get("type", "")
```

After this change we only have to please `mypy` by replacing the initialization and restoration of the `self._printer_name` property from `None` to an empty string:

```python
self._printer_name = ""
```

belono: > Currently, we'd need to check for the KeyError

After this change, it is no longer needed. We can check for ConfigSyntaxError.

Elias0419: It seems like we need the early return to maintain the behavior where a completely missing printer section raises `ConfigSectionMissingError` inside `printer()`

```python
        if not self._printer_name:
            raise exceptions.ConfigSectionMissingError("printer")
```

I propose we have both 

```python
     def __init__(self) -> None:
        # ...
        self._printer_name = ""   
```

```python
        self._printer_config = config.get("printer", None)
        if self._printer_config is None:
            self._has_loaded = True
            return
        printer_name = self._printer_config.get("type", "")
```

and test for both

```python
def test_config_load_with_empty_printer_section() -> None:
    """Config with an empty printer section should raise ConfigSyntaxError."""
    from escpos import config

    c = config.Config()
    with pytest.raises(escpos.exceptions.ConfigSyntaxError):
        c.load_yaml_string("printer: {}")

def test_config_load_yaml_string_without_printer_section() -> None:
    """Config without a printer section should raise ConfigSectionMissingError."""
    from escpos import config

    c = config.Config()
    c.load_yaml_string("{}")
    with pytest.raises(escpos.exceptions.ConfigSectionMissingError):
        c.printer()
```

belono: Looks solid. This is the new suggestion based on your comment:

```suggestion
        self._printer_config = config.get("printer", None)
        if self._printer_config is None:
            self._has_loaded = True
            return
        printer_name = self._printer_config.get("type", "")
```
 Please, modify `self._printer_name = None` to `self._printer_name = ""` in `__init__()` and `_reset_config()` as I can't do it. :love_you_gesture:

Elias0419: Thanks for your help @belono.

Elias0419: Done in [df166d2](https://github.com/python-escpos/python-escpos/pull/707/commits/df166d2cb9a48c7b1432e6a7ef438385c5650ae6) and [a8ebebb](https://github.com/python-escpos/python-escpos/pull/707/commits/a8ebebbc93549e6f303b4e3cf5dea1e6f1df3281)

belono (approved):
It's been a pleasure @Elias0419 . 

Very good job!! :1st_place_medal:

PR #709, opened 2025-10-02
Improve and extend the `_read()` method availability for its users and related methods. Part 1: Cupsprinter, LP and on/offline.

### LINKED ISSUE #611: Introduce IntEnum or similar for printer states
The reason I didn't suggest RT_MASK_ONLINE is that its name is not consistent with the meaning of the returning value. According to [Esc/Pos](https://download4.epson.biz/sec_pubs/pos/reference_en/escpos/dle_eot.html), the int value 8 is for **Offline** status.

We could modify that constant and add a new one for offline:
```python
RT_MASK_ONLINE: int = 0
RT_MASK_OFFLINE: int = 8
```
Then we could do:
```python
if not state or state in [4, 5]:
    return RT_MASK_OFFLINE  # offline
return RT_MASK_ONLINE  # online
```

_Originally posted by @belono in https://github.com/python-escpos/python-escpos/pull/607#discussion_r1428892456_

belono:
### Description

This starts a PR series to improve and extend the availability of the `_read()` method to some connectors where it is currently not implemented.

The objectives of this series are:
- Improve the way we validate the printer's response after a real time status query to reduce false positives. This will also allow for the eventual addition of new status queries (Offline cause, ink status, etc.).
- Implement the unsupported `_read()` method in the following connectors:
  - `File()`
  - `Dummy()`
  - `LP()`
- Extend testing of the related methods.

#### Part 1: Improve on/offline masks consistency and its usage by the Cups and LP connectors:

This will close #611  by:
- Add a `RT_MASK_OFFLINE` constant and change the value of `RT_MASK_ONLINE` to a value that is more consistent with the ESC/POS real-time status response format for both constants (a validator/checker is included in a follow up PR, see #711 ). The new values are also consistent with the rest of RT_MASK_* constants.
- Make use of the on/offline mask values as the return values of the `_read()` method in the `CupsPrinter()`  and `LP()` connectors to improve readability.
- Adapt the related methods and tests to the new values.

### Tested with
TM-U210 Parallel port printer


belono:
Hey there @patkan .

I know you've been short on time lately, but if you could take a look at these 4 requests and merge them, it would help me write the tests. :pray: 

Thanks in advance!

patkan (approved):

PR #711, opened 2025-10-04
Improve and extend the _read() method availability for its users and related methods. Part 2: Response format validator

belono:
### Description
This is part 2 of the series to improve and extend the availability of the _read() method.

#### Part 2: Add a simple response format validator when querying for status:

As stated at the Epson ESC/POS [reference guide](https://download4.epson.biz/sec_pubs/pos/reference_en/escpos/dle_eot.html):

> Each status consists of 1 byte, and the value is 0xx1xx10b.
> The real time status can be differentiated by the bits 0, 1, 4, and 7 from other transmission data, except for data in block data (Header – NUL).

So a status response consists of a byte which bits 0 and 7 are fixed to 0, and bits 1 and 4 are fixed to 1. The remaining bits contain the response to the query according to the tables given in the aforementioned reference guide.

This PR tries to reduce the possibility of false positives when interpreting data from the printer buffer as status responses by checking that the data read is at least in the right response format:
- Add a simple validator method to validate every status format before return it as a response.
- Log an error if validation fails.
- Optionally raise a (new) `ValidationError` exception if validation fails, or return an empty response (default).
- Treat an empty response as a negative response for the `query_status()` users: `is_online()` and `paper_status()` methods.
- Add tests for the status read functions.

### Tested with
TM-U210 Parallel port printer


src/escpos/escpos.py
```diff
@@ -1421,26 +1424,55 @@ def panel_buttons(self, enable: bool = True) -> None:
...
             - RT_STATUS_PAPER: Paper sensor.
+        :param raise_not_valid: Default False.
+                                False to log error but do not raise exception.
+        :raises: :py:exc:`~escpos.exceptions.ValidationError`
         """
         self._raw(mode)
         status = self._read()
+        is_valid = self._check_valid_response(status)
+        if not is_valid:
+            logging.error("Invalid status data: Could't get a valid printer response.")
```
patkan:
```suggestion
            logging.error("Invalid status data: Couldn't get a valid printer response.")
```

patkan (approved):

PR #733, opened 2026-05-17
fix: Update __exit__ method to use Optional types

AustinZhang1024:
### Description

From the newest document, better to allow `None` in the type hint.

This will fix the [invalid-context-manager](https://docs.astral.sh/ty/reference/rules/#invalid-context-manager) error of ty.

> If the context was exited without an exception, all three arguments will be [None](https://docs.python.org/3/library/constants.html#None).

### Tested with
Windows 11


belono (approved):
Hi @AustinZhang1024 !

From [contextmanager.\_\_exit\_\_(exc_type, exc_val, exc_tb)](https://docs.python.org/3.16/library/stdtypes.html#contextmanager.__exit__)
> If an exception occurred while executing the body of the [with](https://docs.python.org/3.16/reference/compound_stmts.html#with) statement, the arguments contain the exception type, value and traceback information. Otherwise, all three arguments are `None`.

You are right, if no exception is raised when exiting a context manager then all three arguments will be `None`, so their types should be `Optional`.

Thanks!

PR #734, opened 2026-05-25
docs: clarify block_text font parameter

### LINKED ISSUE #731: block_text does not set font on POS-5890
The command `block_text()` ignores the `font` setting when printing to the POS-5890. I have to first use `set()` to change the font.

Mirochill:
### Description
Clarifies the `block_text()` documentation so the `font` parameter is described as a column-width lookup hint when `columns` is omitted, not as a command that switches the active printer font. The changelog also notes the documentation clarification.

Closes #731

### Tested with
Not applicable; documentation-only change.

Validation:
- `git diff --check HEAD~1..HEAD`
- Static review of `src/escpos/escpos.py` and `CHANGELOG.rst`
- Tests not run locally


belono (approved):
Hi @Mirochill !

This clarification is very welcome and necessary, as seen in #731 .

You can remove the "Draft" label and leave the pull request open for @patkan to review and integrate.

Thanks!

PR #738, opened 2026-07-20
Drop dependency on six and mock

tiltingpenguin:
### Description
Drop usage of six and switch mock over to the standardlibrary implementation

### Tested with
Only ran the testsuite


src/escpos/escpos.py
```diff
@@ -1135,12 +1131,8 @@ def set(
...
         if underline is not None:
             self._raw(TXT_STYLE["underline"][underline])
-        if font is not None and font != self._font:
-            self._raw(SET_FONT(six.int2byte(self.profile.get_font(font))))
-            self._font = font
-            # Force a fresh code page selection as required by some printer
-            # models (confirmed: NT-5890K).
-            self.magic.reset_encoding()
+        if font is not None:
+            self._raw(SET_FONT(self.profile.get_font(font).to_bytes()))
```
patkan, line 1135:
```suggestion
            self._raw(SET_FONT(self.profile.get_font(font).to_bytes()))
            self._font = font
            # Force a fresh code page selection as required by some printer
            # models (confirmed: NT-5890K).
            self.magic.reset_encoding()
```

src/escpos/escpos.py
```diff
@@ -1135,8 +1131,8 @@ def set(
             self._raw(TXT_STYLE["bold"][bold])
         if underline is not None:
             self._raw(TXT_STYLE["underline"][underline])
-        if font is not None and font != self._font:
-            self._raw(SET_FONT(six.int2byte(self.profile.get_font(font))))
+        if font is not None:
```
patkan:
```suggestion
        if font is not None and font != self._font:
```

PR #764, opened 2026-08-15
tox: pass the args to pytest to select tests

saper:
The following invocation:
    
     tox --conf pyproject.toml run -- -k magicencode
    
will run only magicencode tests


belono (approved):
It works perfectly and doesn't break anything.
Thanks @saper !!

PR #770, opened 2026-09-06
Win32Raw: Add Windows Shared Printer compatibility

DeimElias:
### Description

With this change, it is now possible to print to Windows Shared Printers (r"\\host\printer")

### Tested with

Windows 10 and 11


patkan (approved):

